PR #44, opened 2019-09-30
Fix links

### LINKED ISSUE #37: Fix links
links didn't survive the move into Github. Fix them so that they lead the user to the appropriate sections.

lavaleri:
*Issue #, if available:* Closes #37 

*Description of changes:* Fixes internal references. Linking in this way, especially with the documents in separate directories, makes the raw text more clunky to read. This could be fixed somewhat if we moved to using <a> tags, however this would require significantly more work to shape up.


data-format/message-body.md
```diff
@@ -163,23 +163,23 @@ The Final Frame Sequence number MUST be equal to the total number of frames in t
...
 
 ##### Authentication Tag
 
 The authentication value for the final frame.    
 It is used to authenticate the final frame.  
 The authentication tag length MUST be equal to the authentication tag length of the algorithm suite 
-specified by the [Algorithm Suite ID](#message-header.md#algorithm-suite-id) field.
+specified by the [Algorithm Suite ID](message-header.md#algorithm-suite-id) field.
 
 ## Test Vectors
```
ttjsu-aws: The TODO link on line 186 needs to be fixed, it seems to be point to the https://github.com/awslabs/aws-crypto-tools-test-vector-framework.

PR #46, opened 2019-10-11
add master key and master key provider interface specs

mattsb42-aws:
*Description of changes:*

add master key and master key provider interface specs

### DIFF
--- a/framework/master-key-interface.md
+++ b/framework/master-key-interface.md
@@ -58,6 +58,8 @@ A master key MUST supply itself and MUST NOT supply any other master keys.
 
 Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
 
+A master key SHOULD attempt to decrypt data key using itself.
+
 A master key MUST not attempt to use any other master keys.
 
 ### Generate Data Key
@@ -85,12 +87,18 @@ This interface is used to encrypt a data key.
 
 The master key MUST encrypt the data key.
 
-Inputs the this interface MUST include
+Inputs to this interface MUST include
 the algorithm suite,
 the encryption context,
 the encrypted data key,
 and information that can identify which master key
 was used to encrypt the data key.
+
+The output of this interface MUST include
+a value that this master key can use to obtain
+the plaintext data key.
+Most commonly,
+this will be the result of an encryption operation.
 
 If the master key cannot encrypt the data key,
 the call MUST result in an error.
--- a/framework/master-key-interface.md
+++ b/framework/master-key-interface.md
@@ -1,0 +1,96 @@
+[//]: # (Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.)
+[//]: # (SPDX-License-Identifier: CC-BY-SA-4.0)
+
+# Master Key Interface
+
+## Version
+
+0.1.0-preview
+
+## Implementations
+
+- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/MasterKey.java)
+- [Python](https://github.com/aws/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/key_providers/base.py)
+
+## Overview
+
+Master keys are master key providers that only provide themselves.
+They also support generating and encrypting data keys.
+
+### Legacy
+
+This is a legacy specification.
+Master keys SHOULD NOT be included in any additional implementations.
+Any new implementations MUST include [Keyrings](./keyring-interface.md) instead.
+
+### Consistency
+
+This specification defines the common behavior between the two implementations
+that determine the REQUIRED functionality.
+
+Other specifics of behavior and API vary between the two implementations.
+
+## Definitions
+
+### Conventions used in this document
+
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted
+as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Interface
+
+### Get Master Key
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST supply itself and MUST NOT supply any other master keys.
+
+### Get Master Keys For Encryption
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST supply itself and MUST NOT supply any other master keys.
+
+### Decrypt Data Key
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST not attempt to use any other master keys than itself.
+
+### Generate Data Key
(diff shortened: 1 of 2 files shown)


framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST supply itself and MUST NOT supply any other master keys.
+
+### Decrypt Data Key
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST not attempt to use any other master keys than itself.
```
ajw-aws: I think this is clearer if we remove "than itself" from the end of the sentence

framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Generate Data Key
+
+This interface is used to generate and encrypt a data key.
+
+The master key MUST generate a data key and MUST then encrypt that data key.
+
+Inputs to this interface MUST include
+the algorithm suite
+and the encryption context.
```
ajw-aws: A period at the end, yes or no? Either way, let's be consistent

mattsb42-aws: ? There is a period at the end of the rendered line.

framework/master-key-interface.md
mattsb42-aws: sure

framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+If the master key cannot generate or encrypt the data key,
+the call MUST result in an error.
+
+### Encrypt Data Key
+
+This interface is used to encrypt a data key.
+
+The master key MUST encrypt the data key.
+
+Inputs the this interface MUST include
```
lavaleri: s/the/to/

framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+was used to generate and encrypt the data key.
+
+If the master key cannot generate or encrypt the data key,
+the call MUST result in an error.
+
+### Encrypt Data Key
+
+This interface is used to encrypt a data key.
+
+The master key MUST encrypt the data key.
```
lavaleri: What does this mean. Can we specify what this returns?

framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+Inputs to this interface MUST include
+the algorithm suite
+and the encryption context.
+
+The output of this interface MUST include
+the plaintext data key,
+the data key encrypted under the master key,
+and information that can identify which master key
+was used to generate and encrypt the data key.
```
lavaleri: Is this not the provider ID and key ID?

framework/master-key-interface.md
```diff
@@ -0,0 +1,96 @@
...
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST supply itself and MUST NOT supply any other master keys.
+
+### Decrypt Data Key
+
+Inputs and outputs are the same as for [master key providers](./master-key-provider-interface.md).
+
+A master key MUST not attempt to use any other master keys.
```
lavaleri: A master key MUST attempt to decrypt using itself

framework/master-key-provider-interface.md
```diff
@@ -0,0 +1,106 @@
...
+
+The output of this interface MUST indicate which one of those master keys
+MUST be used to generate the data key.
+
+### Decrypt Data Key
+
+This interface is used to decrypt a data key.
+
+The master key provider SHOULD attempt to decrypt the data key
+using any master keys that it has access to.
```
lavaleri: Using in this case is just using master keys' `Decrypt Data Key`, or is there a need to be more general here?

framework/master-key-provider-interface.md
```diff
@@ -0,0 +1,106 @@
...
+    This concept is equivalent to "key namespace" for Keyrings.
+* Key ID : A value that identifies a master key
+    within the context of a master key provider.
+    This concept is equivalent to "key name" for Keyrings.
+* Provider Info : The value that is written to a serialized encrypted data key
+    that identifies a master key within the context of a master key provider.
+    This MUST always be equal to the master key's key ID
+    with the exception of the raw AES master key.
+    For a detailed description of this exception,
+    see the [Raw AES Keyring specification](./raw-aes-keyring.md).
```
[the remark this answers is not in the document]
mattsb42-aws: I think that mastering it in the Raw Keyring spec is the right thing. I'm on the fence about even doing a Raw master key/provider spec, but if we do, I think it should reference the Raw Keyring spec for any shared behaviors.

framework/master-key-provider-interface.md
```diff
@@ -0,0 +1,106 @@
...
+
+The output of this interface MUST be a master key.
+
+If the master key provider cannot locate an appropriate master key,
+the call MUST result in an error.
+
+### Get Master Keys For Encryption
+
+This interface provides a way for a master key provider to indicate which master keys
+SHOULD be used for encryption.
```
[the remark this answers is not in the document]
mattsb42-aws: This is an artifact of the evolution of these two clients. Originally this the logic for what to do with the returned MKs was in the client itself, so it would have been a "will". Then, when we made CMMs, we moved the logic into the `DefaultCryptoMaterialsManager`. IMO, this makes it a "SHOULD", because the intention is that a MKP will return all of the MKs that it wants you to use, and the only way that we support using MKs is via the `DefaultCryptoMaterialsManager` (pending `MasterKeyProviderKeyring` spec proposal coming soon).

framework/master-key-interface.md
mattsb42-aws: eeeehh....

This is one of the blurry edges of this spec.
The Java client returns a pointer to the object; the Python client returns the provider ID and provider info.

framework/master-key-interface.md
mattsb42-aws: SHOULD, but yes, we should specify that.

framework/master-key-provider-interface.md
```diff
@@ -0,0 +1,106 @@
...
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted
+as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+### Terms
+
+* Provider ID : A value that identifies a master key provider.
+    This concept is equivalent to "key namespace" for Keyrings.
```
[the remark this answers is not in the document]
mattsb42-aws: I think we should define them, yes.

Looking at `message-header.md`, it looks like we punted on saying what the values actually are, and just gave them names.

I think it is worth defining (maybe in the `Terms` section for the keyring spec?) these concepts, even though they are not part of the keyring API.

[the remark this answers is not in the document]
lavaleri: I'm still not sure what the definition of those terms will look like in the interface, but I agree that we should attempt something in the keyring spec, especially if we'd like to be able to compare them in a useful way against MKPs here. #47

PR #48, opened 2019-10-25
Clarify Message Body AAD Content Length field for framed data

### LINKED ISSUE #38: Fix MessageBodyAAD ContentLength
The description for MessageBodyAAD ContentLength doesn't describe how that field is used for framed data. Additionally, update encrypt.md and decrypt.md to accurately describe what should be serialized into this field.

lavaleri:
*Issue #, if available:* closes #38 

*Description of changes:* Updates ContentLength to specify correct behavior in case framed data is being used.

### DIFF
--- a/data-format/message-body-aad.md
+++ b/data-format/message-body-aad.md
@@ -62,8 +62,9 @@ More specifically, depending on the [content type](message-header.md#content-typ
   of the plaintext being encrypted in this frame.
   - For [regular frames](message-body.md#regular-frame), this value MUST equal the value of
     the [frame length](message-hedaer.md#frame-length) field in the message header.
-  - For the [final frame](message-body.md#final-frame), this value MUST be equal to or less than
-    the value of the [frame length](message-hedaer.md#frame-length) field in the message header.
+  - For the [final frame](message-body.md#final-frame), this value MUST be greater than or equal to
+    0 and less than or equal to the value of the [frame length](message-hedaer.md#frame-length)
+    field in the message header.
 
 ## Security Considerations
 
--- a/data-format/message-body-aad.md
+++ b/data-format/message-body-aad.md
@@ -52,12 +52,18 @@ For [non-framed data](#message-body.md#framed-data), the value of this field MUS
 
 ### Content Length
 
-This value depends on the [content type](message-header.md#content-type) of the [message](message.md):
+The length, in bytes, of the plaintext data being encrypted that this message body AAD is associated with.
+
+More specifically, depending on the [content type](message-header.md#content-type) of the [message](message.md):
 
 - For [non-framed data](message-body.md#non-framed-data), this value MUST equal the length, in bytes,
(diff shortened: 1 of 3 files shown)


data-format/message-body-aad.md
```diff
@@ -52,7 +52,12 @@ For [non-framed data](#message-body.md#framed-data), the value of this field MUS
 
 ### Content Length
```
mattsb42-aws: There's actually a third case: final frames.

The reason we went with the wording of "the length of the plaintext data provided to the algorithm is because:

1. For non-framed, it is length of the entire message because it is encrypted in a single call to the encryption algorithm.
1. For standard frames in framed messages, it is the length of the plaintext in the frame, which equal to the message frame length.
1. For final frames in framed messages, it is the length of the plaintext in the frame, which is something greater than or equal to 0 and less than the message frame length.

In each of these cases, the value is the length of the plaintext data provided to the encryption algorithm for which this AAD is constructed.

lavaleri: Gotcha, I will update to make each case explicitly clear. Additionally, I found an incorrect statement in message-body.md regarding frame length so I will clean that up and audit the data format files to make sure we are using `Frame Length` consistently. Thanks!

lavaleri: Updated. Note that for (3) it is actually "greater than or equal to 0 and less than or equal to the message frame length"

data-format/message-body-aad.md
```diff
@@ -52,7 +52,18 @@ For [non-framed data](#message-body.md#framed-data), the value of this field MUS
...
+
+More specifically, depending on the [content type](message-header.md#content-type) of the [message](message.md):
+
+- For [non-framed data](message-body.md#non-framed-data), this value MUST equal the length, in bytes,
+  of the plaintext data provided to the algorithm for encryption.
+- For [framed data](message-body.md#framed-data), this value MUST equal the length, in bytes,
+  of the plaintext being encrypted in this frame.
+  - For [regular frames](message-body.md#regular-frame), this value MUST equal the value of
+    the [frame length](message-hedaer.md#frame-length) field in the message header.
+  - For the [final frame](message-body.md#final-frame), this value MUST be equal to or less than
```
seebees: I think that "greater than or equal to 0 and less than or equal to the message frame length" is better here.

data-format/message-body.md
```diff
@@ -128,7 +128,7 @@ specified by the [Algorithm Suite ID](#message-header.md#algorithm-suite-id) fie
 
 The last frame in a framed body MUST be a Final Frame.  
 
-The length of the Final Frame MUST be less than or equal to the [Frame Length](#message-header.md#frame-length).
+The length of the plaintext to be encrypted in the Final Frame MUST be less than or equal to the [Frame Length](#message-header.md#frame-length).
```
seebees: Sorry, given that you were swayed to add the `MUST be greater than or equal to 0` part to the AAD part, I would add it here as well.

Issue #49, opened 2019-10-29
Define behavior for a KMS-Keyring when given an generator which is an alias


MatthewBennington:
Presently the spec [OnEncrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md#onencrypt) section of the spec says that the keyring should use a "client that calls the AWS region specified in the generator ARN".

In cases where the keyring's generator is an alias (e.g. "alias/MyCryptoKey"), there is no way to determine the region to make call is.

seebees:
This is also true for the keyIds as well.  Notes are in the PR, but this is not a generator specific problem

acioc:
Closing in favor of https://github.com/awslabs/aws-encryption-sdk-specification/issues/165

PR #50, opened 2019-10-29
Remove lines mandating region be specified by KMS keyring generator

MatthewBennington:
*Issue #, if available:* #49 

*Description of changes:* Changes to OnEncrypt/OnDecrypt to relax the dependence on generator ARN for AWS region.

*What this PR does NOT cover:* This PR does NOT redefine the Client Supplier to take a CMK ARN as input.

### DIFF
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -209,7 +209,7 @@ as described above, OnEncrypt MUST also attempt to encrypt the plaintext data ke
 
 To attempt to encrypt the plaintext data key with a particular CMK, OnEncrypt MUST call [KMS Encrypt](#kms-encrypt).
 
-For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the CMK ARN, then the
+For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the [Key Id](#key-ids), then the
 [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the client returned by the
 [client supplier](#client-supplier) when given that region as input.
 If an AWS region cannot be extracted from the CMK ARN then the KMS Keyring MUST input a value denoting an unknown region.
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -160,9 +160,9 @@ If the input [encryption materials](#structures.md#encryption-materials) do not 
 and this keyring has a [generator](#generator) defined, and onEncrypt MUST attempt to generate a new plaintext data key
 and encrypt that data key by calling [KMS GenerateDataKey](#kms-generatedatakey).
 
-The [KMS client](#kms-client) that calls [KMS GenerateDataKey](#kms-generatedatakey) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
+[KMS GenerateDataKey](#kms-generatedatakey) MUST be the client returned by the [client supplier](#client-supplier)
+when given that region as input.
 If the [client supplier](#client-supplier) does not provide any client for the given region for this GenerateDataKey call,
 OnEncrypt MUST fail.
 
@@ -266,9 +266,10 @@ in the input encrypted data key list with the following conditions, until it suc
 To attempt to decrypt a particular [encrypted data key](#structures.md#encrypted-data-key),
 OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 
-The [KMS client](#kms-client) that calls [KMS Decrypt](#kms-decrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
+[KMS Decrypt](#kms-decrypt) MUST be the client returned by the [client supplier](#client-supplier)
+when given that region as input.
 If the [client supplier](#client-supplier) does not provide any client for the given region for the Decrypt call,
 OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
 
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -271,7 +271,6 @@ OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 For each [KMS Decrypt](#kms-decrypt) call, if an AWS region can be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information).
 The KMS Keyring MUST call [KMS Decrypt](#kms-decrypt) using the client supplied by the [client supplier](#client-supplier), given the region as input.
 
-If an AWS region cannot be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information) then the KMS Keyring MUST input to the [client supplier](#client-supplier) a value denoting an unknown region.
 If the client supplier does not provide any client for the given region for this Decrypt call, OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
 
 When calling [KMS Decrypt](#kms-decrypt), the keyring MUST call with a request constructed as follows:
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -268,7 +268,7 @@ in the input encrypted data key list with the following conditions, until it suc
 To attempt to decrypt a particular [encrypted data key](#structures.md#encrypted-data-key),
 OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 
-For each [KMS Decrypt](#kms-decrypt) call, if an AWS region can be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information).
+For each [KMS Decrypt](#kms-decrypt) call, an AWS region MUST be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information).
 The KMS Keyring MUST call [KMS Decrypt](#kms-decrypt) using the client supplied by the [client supplier](#client-supplier), given the region as input.
 
 If the client supplier does not provide any client for the given region for this Decrypt call, OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -194,8 +194,8 @@ If verified, OnEncrypt MUST do the following with the response from [KMS Generat
   - the [key provider information](#data-strucures.md#key-provider-information) is the response `KeyId`.
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
-  - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains the value of the KMS Keyring's [generator](#generator) field.
+  - The string field KeyNamespace is "aws-kms".
+  - The string field KeyName is the value of the KMS Keyring's [generator](#generator) field.
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [GENERATED DATA KEY](#structures.md#supported-flags)
     - [ENCRYPTED DATA KEY](#structures.md#supported-flags)
@@ -299,8 +299,8 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
 - set the plaintext data key on the [decryption materials](#structures.md#decryption-materials) as the response `Plaintext`.
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
-  - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains the [encrypted data key's key provider info](structures.md#key-provider-information).
+  - The string field KeyNamespace is "aws-kms".
+  - The string field KeyName is the [encrypted data key's key provider info](structures.md#key-provider-information).
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [DECRYPTED DATA KEY](#structures.md#supported-flags)
     - [VERIFIED ENCRYPTION CONTEXT](#structures.md#supported-flags)
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -211,7 +211,7 @@ To attempt to encrypt the plaintext data key with a particular CMK, OnEncrypt MU
 For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the [Key ID](#key-ids), then the
 [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the client returned by the
 [client supplier](#client-supplier) when given that region as input.
-If an AWS region cannot be extracted from the CMK ARN then the KMS Keyring MUST input a value denoting an unknown region.
+If an AWS region cannot be extracted from the Key ID then the KMS Keyring MUST input a value denoting an unknown region.
 If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call, OnEncrypt MUST skip that particular CMK.
 
 To encrypt the plaintext data key with a CMK, OnEncrypt MUST call [KMS Encrypt](#encrypt) with a request
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -236,8 +236,8 @@ If the call succeeds, OnEncrypt MUST do the following with the response from [KM
     Note that the `KeyId` in the response is always in key ARN format.
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
-  - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains the CMK ARN used for [KMS Encrypt](#kms-encrypt).
+  - The string field KeyNamespace is "aws-kms".
+  - The string field KeyName is the response `KeyId`. Note that the `KeyId` in the response is always in key ARN format.
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [ENCRYPTED DATA KEY](#structures.md#supported-flags)
     - [SIGNED ENCRYPTION CONTEXT](#structures.md#supported-flags)
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -165,7 +165,7 @@ If an AWS region can be extracted from the [generator](#generator), then the [KM
 [KMS GenerateDataKey](#kms-generatedatakey) MUST be the client returned by the [client supplier](#client-supplier)
 when given that region as input.
 If an AWS region cannot be extracted from the [generator](#generator) then the KMS Keyring MUST input a value denoting an unknown region.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this [KMS GenerateDataKey](#kms-generatedatakey) call, OnEncrypt MUST not modify the [encryption materials](#structures.md#encryption-materials.md)
+If the [client supplier](#client-supplier) does not provide any client for the given region for this [KMS GenerateDataKey](#kms-generatedatakey) call, OnEncrypt MUST NOT modify the [encryption materials](#structures.md#encryption-materials.md)
 and MUST fail.
 
 When calling [KMS GenerateDataKey](#kms-generatedatakey), the keyring MUST call with a request constructed as follows:
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -64,9 +64,10 @@ On keyring initialization, a keyring MAY define the following:
 
 ### Client Supplier
 
-A function that maps a region string to a KMS client that can make the following API calls in the given AWS region:
+A function which MUST take as input either a region string (e.g. `us-east-1`), or some value denoting an unknown region,
+and MAY return a KMS Client which can make the following API calls in the given AWS region:
 
-- [GenerateDataKey](#kms-generate-data-key)
+- [GenerateDataKey](#kms-generatedatakey)
 - [Encrypt](#kms-encrypt)
 - [Decrypt](#kms-decrypt)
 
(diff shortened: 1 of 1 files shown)


framework/kms-keyring.md
```diff
@@ -160,9 +160,9 @@ If the input [encryption materials](#structures.md#encryption-materials) do not
 and this keyring has a [generator](#generator) defined, and onEncrypt MUST attempt to generate a new plaintext data key
 and encrypt that data key by calling [KMS GenerateDataKey](#kms-generatedatakey).
 
-The [KMS client](#kms-client) that calls [KMS GenerateDataKey](#kms-generatedatakey) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
```
seebees: This is also true of the KMS Encrypt call below.

framework/kms-keyring.md
```diff
@@ -266,9 +266,10 @@ in the input encrypted data key list with the following conditions, until it suc
 To attempt to decrypt a particular [encrypted data key](#structures.md#encrypted-data-key),
 OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 
-The [KMS client](#kms-client) that calls [KMS Decrypt](#kms-decrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
```
seebees: On the other hand.  For decrypt the CMK _must_ be the ARN.

MatthewBennington: The CMK used for decryption must be an ARN, but that CMK is not necessarily the generator. I don't understand how this comment relates to this line, can you clarify?

lavaleri: For decryption, only keys with a well formed key ARN will produce a successful decryption. This is because the keyring will only attempt to call KMS Decrypt for EDKs where the keyProvider matches a keyID. Because the EDK is serialized to always use the key ARN as the keyProvider, the only way to get the KMS Keyring to call KMS Decrypt for an EDK is if you include the key ARN in the keyIDs/generator.

Thus, if we ever actually get to attempting to call KMS, we know that the key in question has a region. HOWEVER, the generator could still just be an alias... I think it still makes sense to specify the "everything is a well formed ARN" case, and leave specifying other behavior as part of #49

MatthewBennington: > Because the EDK is serialized to always use the key ARN as the keyProvider, the only way to get the KMS Keyring to call KMS Decrypt for an EDK is if you include the key ARN in the keyIDs/generator.

If I understand correctly, then if a given EDK's KeyProvider matches one of the keyring's KeyIDs, then it will attempt to decrypt. In this scenario the keyring would need to create a client by inputting a region into the client supplier. To determine that region it would need to parse the generator string, which may be an alias.

framework/kms-keyring.md
```diff
@@ -160,9 +160,9 @@ If the input [encryption materials](#structures.md#encryption-materials) do not
 and this keyring has a [generator](#generator) defined, and onEncrypt MUST attempt to generate a new plaintext data key
 and encrypt that data key by calling [KMS GenerateDataKey](#kms-generatedatakey).
 
-The [KMS client](#kms-client) that calls [KMS GenerateDataKey](#kms-generatedatakey) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
+[KMS GenerateDataKey](#kms-generatedatakey) MUST be the client returned by the [client supplier](#client-supplier)
+when given that region as input.
```
mattsb42-aws: This feels like it is drawing an odd relationship that I don't think exists. The way this is worded make it sound to me like we are saying that there is something special about the region name extracted from *this* ARN that makes it different from the same region name extracted from *some other* ARN. This is not the case (though, side note, this is why I wanted to the client provider to accept a key name rather than a region name, but that's a different conversation).

I think we can simplify this to something along the lines of the following.

**Client Provider**

* The client provider MUST require a region name.
* The client provider MUST either provide a client for the requested region or fail.


**Keyring**

* If the key ID is an ARN then the keyring MUST call the client provider with the region in the ARN.
* If the key ID is *not* an ARN then the keyring MAY select a region with which to call the client provider.

MatthewBennington: > The way this is worded make it sound to me like we are saying that there is something special about the region name extracted from this ARN that makes it different from the same region name extracted from some other ARN.

As I understand the spec, there is something special about this region. No other region derived from an ARN determines the region used by the client (or after this PR, the region given to the client supplier).

mattsb42-aws: There is something special about the generator key. That generator key happens to exist in a single region. The region is a side effect.

MatthewBennington: How is this?

> If the generator is formatted as either a well-formed key ARN, or a well-formed alias ARN, then an AWS region MUST be derived from the generator and used to call the client supplier. The client produced as output of the client supplier MUST be used to call KMS Encrypt/Decrypt/GenerateDataKey.

seebees: >The client provider MUST require a region name.

I disagree.
The AWS SDK can pick up a default region in the same way it gets credentials.
The client provider is the interface to the AWS SDK.
Let the SDK handle it.
If there is any additional complication regarding what a default region should be,
putting this logic in a function is much easier for customers.

I have an example of what I'm thinking about here: https://github.com/aws/aws-encryption-sdk-javascript/pull/231

framework/kms-keyring.md
```diff
@@ -179,8 +180,7 @@ If the call to [KMS GenerateDataKey](#kms-generatedatakey) does not succeed, OnE
 modify the [encryption materials](#structures.md#encryption-materials) and MUST fail.
 
 If the call succeeds, OnEncrypt MUST verify that the response `Plaintext` length matches the specification
-of the [algorithm suite](#algorithm-suites.md).
-If it does not, OnEncrypt MUST fail.
+of the [algorithm suite](algorithm-suites.md)'s Key Derivation Input Length field. If it does not, OnEncrypt MUST fail.
```
lavaleri: Here and elsewhere: newline after `.`

framework/kms-keyring.md
```diff
@@ -193,23 +193,23 @@ If verified, OnEncrypt MUST do the following with the response from [KMS Generat
...
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
   - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains [generator](#generator).
+  - The string field KeyName contains the value of the KMS Keyring's [generator](#generator) field.
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [GENERATED DATA KEY](#structures.md#supported-flags)
     - [ENCRYPTED DATA KEY](#structures.md#supported-flags)
     - [SIGNED ENCRYPTION CONTEXT](#structures.md#supported-flags)
 
-Given a plaintext data key in the [encryption materials](#structures.md#encryption-materials),
```
lavaleri: Without this qualifier is it clear what `plaintext data key` is referring to on the next line? Maybe there is a way to restate this which is less confusing, or do you think it's not necessary?

framework/kms-keyring.md
```diff
@@ -64,7 +64,7 @@ On keyring initialization, a keyring MAY define the following:
 
 ### Client Supplier
 
-A function that returns a KMS client that can make the following API calls in a particular AWS region:
+A function that maps a region string to a KMS client that can make the following API calls in the given AWS region:
```
lavaleri: I think we want to be clear what the inputs/outputs are. The input is a string, the output is a KMS client.

framework/kms-keyring.md
```diff
@@ -160,9 +160,10 @@ If the input [encryption materials](#structures.md#encryption-materials) do not
 and this keyring has a [generator](#generator) defined, and onEncrypt MUST attempt to generate a new plaintext data key
 and encrypt that data key by calling [KMS GenerateDataKey](#kms-generatedatakey).
 
-The [KMS client](#kms-client) that calls [KMS GenerateDataKey](#kms-generatedatakey) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
+[KMS GenerateDataKey](#kms-generatedatakey) MUST be the client returned by the [client supplier](#client-supplier)
+when given that region as input. If an AWS region cannot be extraced from the [generator](#generator),
+then the KMS Keyring MAY choose an arbitrary region.
```
lavaleri: We might want to clarify what we mean by `May choose an arbitrary region`. What we really mean is that if you can't get the region from the key, then the KMS keyring is allowed to use a KMS client from any arbitrary region to make the call to KMS. Maybe just say "If the AWS region cannot be extracted...the KMS Keyring MAY use a KMS client from any region"

framework/kms-keyring.md
```diff
@@ -266,9 +266,12 @@ in the input encrypted data key list with the following conditions, until it suc
 To attempt to decrypt a particular [encrypted data key](#structures.md#encrypted-data-key),
 OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 
-The [KMS client](#kms-client) that calls [KMS Decrypt](#kms-decrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
+
+If an AWS region can be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s
```
lavaleri: An AWS region MUST be extractable from the EDK. This is because the EDK key provider info serializes the keyARN, and never the alias. If the KMS Keyring cannot extract the region in this case, it should result in an error.

framework/kms-keyring.md
MatthewBennington: Good point about clarifying `plaintext data key`. I removed this line because I think it creates the appearance of an else to the if on line 154. I will try to find wording that addresses this.

framework/kms-keyring.md
mattsb42-aws: This is where I hit a bit of a sticking point: Who is the actor here?

# Region discernible from key ID

1. The KMS Keyring *through some undefined mechanism* attempts to determine region from key ID.
1. The KMS Keyring passes the region name to the client supplier.
1. The client supplier provides the KMS Keyring with a client.
1. The KMS Keyring uses that client.

# Region not discernible from key ID

## Keyring is the actor

1. The KMS Keyring *through some undefined mechanism* attempts to determine region from key ID.
1. The KMS Keyring fails to determine region from key ID.
1. The KMS Keyring *through some undefined mechanism* obtains a client.

## Client Supplier is the actor

1. The KMS Keyring *through some undefined mechanism* attempts to determine region from key ID.
1. The KMS Keyring fails to determine region from key ID.
1. The KMS Keyring tells the Client Supplier that it doesn't know what the region is.
1. The Client Supplier MAY provide the KMS Keyring with a client.
1. The KMS Keyring uses that client or fails if the Client Supplier fails to provide one.

lavaleri: There are two points being brought up here:
1. We should specify that the client supplier is responsible for determining default/"I can't get a region" behavior instead of the keyring. As said above, this makes sense because otherwise we violate the API contract.
2. In the view of a client supplier, it can be a bit tricky what the default behavior SHOULD be. Generally the supplier should make use of the AWS SDK defualt region, however the "limit regions" supplier example above shows how this might not always be the desired behavior. I think more precisely defining the restrictions on client supplier behavior is out of scope for this PR.

As such, I think these are the next steps for this PR:

- Update client supplier interface to say that it MUST support some input (i.e. empty string) to indicate that the region in unknown
- Update L66 to say that if the region can't be determined, then the Keyring MUST input the "I don't know" input to the client supplier, and if the client supplier returns a client the Keyring MUST use that client.

@mattsb42-aws @seebees Thoughts?

seebees: Good to me.

mattsb42-aws: Sounds good!

framework/kms-keyring.md
MatthewBennington: Can you help me understand in what way is this sentence unclear about that?

framework/kms-keyring.md
MatthewBennington: As I read the spec, the Client Supplier is an interface which we let the customer implement. As such, I don't think it makes sense to define default behavior.

MatthewBennington: How about:

> ### Client Supplier
> A function which MUST take as input either a region string (e.g. `us-east-1`), or some value denoting an unknown region, and MAY return a KMS Client which can make the following API calls in the given AWS region:

and

> If an AWS region cannot be extracted from the generator then the KMS Keyring MUST input a value denoting an unknown region. If the client supplier does not provide any client for the given region for this Encrypt call, OnEncrypt MUST skip that particular CMK.

I've opted not to define the value indicating an unknown region, as that seems fairly language specific.

mattsb42-aws: `s/which/that/` but otherwise LGTM @MatthewBennington

framework/kms-keyring.md
```diff
@@ -266,14 +268,11 @@ in the input encrypted data key list with the following conditions, until it suc
...
 
-If an AWS region can be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s
-[key provider info](structures.md#key-provider-information), then the [KMS client](#kms-client) that calls
-[KMS Decrypt](#kms-decrypt) MUST be the client returned by the [client supplier](#client-supplier)
-when given that region as input. If an AWS region cannot be extraced from the CMK ARN, then the
-KMS Keyring MAY choose an arbitrary region.
-If the [client supplier](#client-supplier) does not provide any client for the given region for the Decrypt call,
-OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
+If an AWS region cannot be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information) then the KMS Keyring MUST input to the [client supplier](#client-supplier) a value denoting an unknown region.
+If the client supplier does not provide any client for the given region for this Decrypt call, OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
```
lavaleri: We shouldn't specify this for OnDecrypt, as the specified behavior above will never allow us to try to obtain a region from a key that doesn't have a region. Instead, the spec at this point should state that you MUST get the region from the key. If for whatever the implementation is unable to get a region from the key at this point, something is really wrong, and the implementation should fail due to being unable to follow the specification.

framework/kms-keyring.md
lavaleri: new wording LGTM

framework/kms-keyring.md
lavaleri: The latest wording addresses my concern. I was just pointing out that using `maps` isn't as clear as explicitly stating what the input is and what the output is.

framework/kms-keyring.md
MatthewBennington: Good catch

framework/kms-keyring.md
```diff
@@ -266,11 +268,10 @@ in the input encrypted data key list with the following conditions, until it suc
 To attempt to decrypt a particular [encrypted data key](#structures.md#encrypted-data-key),
 OnDecrypt MUST call [KMS Decrypt](#kms-decrypt).
 
-The [KMS client](#kms-client) that calls [KMS Decrypt](#kms-decrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
-If the [client supplier](#client-supplier) does not provide any client for the given region for the Decrypt call,
-OnDecrypt MUST skip that particular [encrypted data key](#encrypted-data-key).
+For each [KMS Decrypt](#kms-decrypt) call, if an AWS region can be extracted from the [encrypted data key](structures.md#encrypted-data-key)'s [key provider info](structures.md#key-provider-information).
+The KMS Keyring MUST call [KMS Decrypt](#kms-decrypt) using the client supplied by the [client supplier](#client-supplier), given the region as input.
```
lavaleri: Whoops, this wording needs to also be updated to remove the "if an AWS region can be extracted".

framework/kms-keyring.md
```diff
@@ -207,11 +209,11 @@ as described above, OnEncrypt MUST also attempt to encrypt the plaintext data ke
 
 To attempt to encrypt the plaintext data key with a particular CMK, OnEncrypt MUST call [KMS Encrypt](#kms-encrypt).
 
-The [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call,
-OnEncrypt MUST skip that particular CMK.
+For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the CMK ARN, then the
```
seebees:
```suggestion
For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the [Key Id](#key-ids), then the
```
Part of what we are doing here is dealing with things that are not ARNs

framework/kms-keyring.md
```diff
@@ -235,7 +237,7 @@ If the call succeeds, OnEncrypt MUST do the following with the response from [KM
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
   - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains [generator](#generator).
+  - The string field KeyName contains the CMK ARN used for [KMS Encrypt](#kms-encrypt).
```
seebees: Both the C and JS ESDKs do not do this.  They both use the value returned from KMS.

seebees:
I also note that there is a conflict between the things we take as CMKs and what we send to KMS.
We state that we support alias, ARN, and alias ARN.  But the clientSupplier will be passed a form of unknown if a region can not be extracted.
While I feel strongly that configuring the KMS Keyring with a raw key id is a bad operational idea, it would work.

framework/kms-keyring.md
```diff
@@ -81,7 +81,7 @@ Each Key ID MUST be one of the following:
 
 - A CMK [alias](https://docs.aws.amazon.com/kms/latest/developerguide/programming-aliases.html) (e.g. "alias/MyCryptoKey")
 - A well-formed key ARN (e.g. arn:aws:kms:us-east-1:999999999999:key/01234567-89ab-cdef-fedc-ba9876543210)
-- A well-formned alias ARN (e.g. arn:aws:kms:us-east-1:999999999999:alias/MyCryptoKey)
+- A well-formed alias ARN (e.g. arn:aws:kms:us-east-1:999999999999:alias/MyCryptoKey)
```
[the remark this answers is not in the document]
seebees: From the KMS side, `/` is the _only_ supported version.

framework/kms-keyring.md
```diff
@@ -64,9 +64,10 @@ On keyring initialization, a keyring MAY define the following:
 
 ### Client Supplier
 
-A function that returns a KMS client that can make the following API calls in a particular AWS region:
+A function which MUST take as input either a region string (e.g. `us-east-1`), or some value denoting an unknown region,
```
mattsb42-aws:
```suggestion
A function that MUST take as input either a region string (e.g. `us-east-1`), or some value denoting an unknown region,
```

framework/kms-keyring.md
```diff
@@ -64,9 +64,10 @@ On keyring initialization, a keyring MAY define the following:
...
+and MAY return a KMS Client which can make the following API calls in the given AWS region:
```
mattsb42-aws:
```suggestion
and MAY return a KMS Client that can make the following API calls in the given AWS region:
```

framework/kms-keyring.md
```diff
@@ -160,11 +161,12 @@ If the input [encryption materials](#structures.md#encryption-materials) do not
...
-The [KMS client](#kms-client) that calls [KMS GenerateDataKey](#kms-generatedatakey) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this GenerateDataKey call,
-OnEncrypt MUST fail.
+If an AWS region can be extracted from the [generator](#generator), then the [KMS client](#kms-client) that calls
+[KMS GenerateDataKey](#kms-generatedatakey) MUST be the client returned by the [client supplier](#client-supplier)
+when given that region as input.
+If an AWS region cannot be extracted from the [generator](#generator) then the KMS Keyring MUST input a value denoting an unknown region.
+If the [client supplier](#client-supplier) does not provide any client for the given region for this [KMS GenerateDataKey](#kms-generatedatakey) call, OnEncrypt MUST not modify the [encryption materials](#structures.md#encryption-materials.md)
```
mattsb42-aws:
```suggestion
If the [client supplier](#client-supplier) does not provide any client for the given region for this [KMS GenerateDataKey](#kms-generatedatakey) call, OnEncrypt MUST NOT modify the [encryption materials](#structures.md#encryption-materials.md)
```

framework/kms-keyring.md
```diff
@@ -193,12 +195,12 @@ If verified, OnEncrypt MUST do the following with the response from [KMS Generat
...
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
   - The string field KeyNamespace contains "aws-kms".
-  - The string field KeyName contains [generator](#generator).
+  - The string field KeyName contains the value of the KMS Keyring's [generator](#generator) field.
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [GENERATED DATA KEY](#structures.md#supported-flags)
     - [ENCRYPTED DATA KEY](#structures.md#supported-flags)
     - [SIGNED ENCRYPTION CONTEXT](#structures.md#supported-flags)
-
+
```
mattsb42-aws:
```suggestion

```

framework/kms-keyring.md
```diff
@@ -207,11 +209,11 @@ as described above, OnEncrypt MUST also attempt to encrypt the plaintext data ke
 
 To attempt to encrypt the plaintext data key with a particular CMK, OnEncrypt MUST call [KMS Encrypt](#kms-encrypt).
 
-The [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the
-client returned by the [client supplier](#client-supplier).
-The client MUST be a client that calls the AWS region specified in the [generator](#generator) ARN.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call,
-OnEncrypt MUST skip that particular CMK.
+For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the [Key Id](#key-ids), then the
```
mattsb42-aws:
```suggestion
For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted from the [Key ID](#key-ids), then the
```

framework/kms-keyring.md
```diff
@@ -207,11 +209,11 @@ as described above, OnEncrypt MUST also attempt to encrypt the plaintext data ke
...
+[KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the client returned by the
+[client supplier](#client-supplier) when given that region as input.
+If an AWS region cannot be extracted from the CMK ARN then the KMS Keyring MUST input a value denoting an unknown region.
```
mattsb42-aws:
```suggestion
If an AWS region cannot be extracted from the Key ID then the KMS Keyring MUST input a value denoting an unknown region.
```

framework/kms-keyring.md
```diff
@@ -300,7 +301,7 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
   in the input [encryption materials](#structures.md#encryption-materials), constructed as follows:
   - The string field KeyNamespace contains "aws-kms".
```
mattsb42-aws:
```suggestion
  - The string field KeyNamespace is "aws-kms".
```

framework/kms-keyring.md
```diff
@@ -300,7 +301,7 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
-  - The string field KeyName contains [generator](#generator).
+  - The string field KeyName contains the [encrypted data key's key provider info](structures.md#key-provider-information).
```
mattsb42-aws:
```suggestion
  - The string field KeyName is the [encrypted data key's key provider info](structures.md#key-provider-information).
```

framework/kms-keyring.md
MatthewBennington: "CMK ARN" is more consistent with the rest of the document. "Key ID" elsewhere is only used to refer elements of the "Key IDs" list.

In my opinion we should revise the spec to have better wording around "CMK" vs "CMK ARN" vs "Key ID". But that is not within the scope of this PR.

MatthewBennington: Actually, re-reading through the file, I see the "Key ID" is more consistent within the paragraph. I will add this change.

Issue #53, opened 2019-11-05
Rewrite Keyring/CMM interfaces in functional style


robin-aws:
The changes I made in https://github.com/awslabs/aws-encryption-sdk-dafny/pull/73 need to be ported over to the way the specification is phrased as well. The wording in many places assume that functions like OnEncrypt accept an EncryptionMaterials instance and may or may not mutate fields on it, but the Dafny interfaces now accept individual parameters and return immutable structures with results.

It will be somewhat tricky to map this description to existing ESDK implementations that are in fact operating via side-effects, but it should be possible to draw equivalences.

PR #56, opened 2019-11-20
Rewrite Keyring interfaces in functional style

robin-aws:
https://github.com/awslabs/aws-encryption-sdk-specification/issues/53

The current specification defines the behaviour of the `Keyring` concept in terms of mutation side-effects on the `EncryptionMaterials` and `DecryptionMaterials` structures. This change refactors the `Keyring` interfaces to specify separate immutable input and output structures.

In order to do so, it introduces a new `Data Key Materials` concept that acts as the return value from `Keyring.OnEncrypt` and `Keyring.OnDecrypt`.

### DIFF
--- a/framework/cmm-interface.md
+++ b/framework/cmm-interface.md
@@ -85,17 +85,10 @@ it MUST return [encryption materials](#structures.md#encryption-materials) appro
 
 The encryption materials returned MUST include the following:
 
-- [Algorithm Suite](#algorithm-suite.md)
-    - If the encryption materials request contains an algorithm suite, the encryption materials returned SHOULD contain the same algorithm suite.
-- Plaintext Data Key
-- [Encrypted Data Keys](#structures.md#encrypted-data-keys)
-    - Every encrypted data key in this list MUST correspond to the above plaintext data key. 
+- [Data Key Materials](#structures.md#data-key-materials.md)
+    - If the encryption materials request contains an algorithm suite, the data key materials returned SHOULD contain the same algorithm suite.
 - [Encryption Context](#structures.md#encryption-context)
-    - The CMM MAY modify the encryption context.   
-
-The encryption materials returned MAY include the following:
-
-- [Keyring Trace](#structures.md#keyring-trace)
+    - The CMM MAY return an encryption context that differs from the input.   
 
 If the algorithm suite contains a [signing algorithm](#algorithm-suites.md#signature-algorithm): 
 
--- a/framework/keyring-interface.md
+++ b/framework/keyring-interface.md
@@ -43,9 +43,10 @@ The following inputs are REQUIRED:
 
 The following inputs are OPTIONAL:
 
-- [Plaintest Data Key](#structures.md#plaintext-data-key)
+- [Plaintext Data Key](#structures.md#plaintext-data-key)
 
-This interface MAY return [data key materials](#structures.md#data-key-materials) appropriate for the request. 
+This interface MAY return [data key materials](#structures.md#data-key-materials) appropriate for the request. If it does return this output
+and a plaintext data key was provided as input, the output must specify the same plaintext data key.
 
 The output of this interface is driven by two behaviours:
 
@@ -122,7 +123,8 @@ This interface MAY perform the following behavior:
 
 - [Decrypt data key](#decrypt-data-key)
 
-If this keyring attempted the above behavior, and succeeded, it MUST output the resulting [data key materials](#structures.md#data-key-materials). The [algorithm suite](#algorithm-suites.md) in the result must match the input [algorithm suite](#algorithm-suites.md).
+If this keyring attempted the above behavior, and succeeded, it MUST output the resulting [data key materials](#structures.md#data-key-materials). The [algorithm suite](#algorithm-suites.md) in the result must match the input [algorithm suite](#algorithm-suites.md), and the [encrypted data keys](#structures.md#encrypted-data-keys)
+list must be empty.
 
 If the keyring did not attempt the above behavior, it MUST produce no output.
 
--- a/framework/keyring-interface.md
+++ b/framework/keyring-interface.md
@@ -36,37 +36,47 @@ in this document are to be interpreted as described in [RFC 2119](https://tools.
 
 ### OnEncrypt
 
-This interface takes [encryption materials](#structures.md#encryption-materials) as input
-and MAY modify it with any of the following behaviors:
+The following inputs are REQUIRED:
+
+- [Algorithm Suite](#algorithm-suites.md)
+- [Encryption Context](#encryption-context)
+
+The following inputs are OPTIONAL:
+
+- [Plaintest Data Key](#structures.md#plaintext-data-key)
+
+This interface MAY return [data key materials](#structures.md#data-key-materials) appropriate for the request. 
+
+The output of this interface is driven by two behaviours:
 
 - [Generate data key](#generate-data-key)
 - [Encrypt data key](#encrypt-data-key)
 
 If this keyring attempted any of the above behaviors, and successfully completed those behaviors,
-it MUST output the modified [encryption materials](#structures.md#encryption-materials).
+it MUST output the result as [data key materials](#structures.md#data-key-materials).
 
-If the keyring did not attempt any of the above behaviors, it MUST output the
-[encryption materials](#structures.md#encryption-materials) unmodified.
+If the keyring did not attempt any of the above behaviors, it MUST produce no output.
 
 #### Generate Data Key
 
-If the [encryption materials](#structures.md#encryption-materials) do not contain a plaintext data key,
+If the input does not include a plaintext data key,
 OnEncrypt MAY generate a data key.
-If the encryption materials contain a plaintext data key, OnEncrypt MUST NOT generate a data key.
+If the input does include a plaintext data key, OnEncrypt MUST NOT generate a data key.
 
-Generate Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Generate Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
 
 - [plaintext data key](#structures.md#plaintext-data-key)
 - [keyring trace](#structures.md#keyring-trace)
 
 To perform this behavior, the keyring generates a [plaintext data key](#structures.md#plaintext-data-key)
-and sets the resulting plaintext data key on the [encryption materials](#structures.md#encryption-materials).
+and includes the resulting plaintext data key in the [data key materials](#structures.md#data-key-materials)' 
+[data key materials](#structures.md#data-key-materials).
 
 The length of the output plaintext data key MUST be equal to the KDF input length of the [algorithm suite](#algorithm-suites.md)
 specified in the [encryption materials](#structures.md#encryption-materials).
 The value of the plaintext data key MUST consist of cryptographically secure (pseudo-)random bits.
 
-If OnEncrypt updates the [encryption materials](#data-structure.md#encryption-materials) with a plaintext data key,
+If OnEncrypt generates a plaintext data key,
 the [keyring trace](#structures.md#keyring-trace) returned by OnEncrypt MUST include one trace
 associated with this keyring that has the [GENERATED DATA KEY](#structures.md#generated-data-key) flag.
 
@@ -75,26 +85,26 @@ perform the [Encrypt Data Key](#encrypt-data-key) behavior.
 
 #### Encrypt Data Key
 
-If the [encryption materials](#structures.md#encryption-materials) contain a plaintext data key,
-OnEncrypt MAY encrypt a data key.
-If the encryption materials do not contain a plaintext data key, OnEncrypt MUST NOT encrypt a data key.
+If the input contains a plaintext data key, or if the keyring generates a data key,
+OnEncrypt MAY encrypt the data key.
+Otherwise, OnEncrypt MUST NOT encrypt a data key.
 
-Encrypt Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Encrypt Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
 
 - [encrypted data keys](#structures.md#encrypted-data-keys)
 - [keyring trace](#structures.md#keyring-trace)
 
 To perform this behavior, the keyring creates one or more [encrypted data keys](#structures.md#encrypted-data-key)
-using the plaintext data key from the [encryption materials](#structures.md#encryption-materials) as input,
-and appends the [encrypted data keys](#structures.md#encrypted-data-key) to the encrypted data key list
-in the [encryption materials](#structures.md#encryption-materials).
+using the plaintext data key as input,
+and includes the [encrypted data keys](#structures.md#encrypted-data-key)
+in the [data key materials](#structures.md#data-key-materials) result.
 
 The [encrypted data keys](#structures.md#encrypted-data-key) produced by this keyring MUST
 have [ciphertexts](#structures.md#ciphertext) that can be decrypted to the plaintext data key in the
-[encryption materials](#structures.md#encryption-materials).
+[data key materials](#structures.md#data-key-materials).
 
-If OnEncrypt updates the [encryption materials](#data-structure.md#encryption-materials) with at least
-one new [encrypted data key](#structures.md#encrypted-data-key),
+If OnEncrypt outputs [data key materials](#structures.md#data-key-materials) with at least
+one [encrypted data key](#structures.md#encrypted-data-key),
 the [keyring trace](#structures.md#keyring-trace) returned by OnEncrypt MUST include at least one trace
 associated with this keyring that has the [ENCRYPTED DATA KEY](#structures.md#encrypted-data-key) flag.
 Note that this trace MAY include more than one flag,
@@ -102,24 +112,23 @@ for example the [SIGNED ENCRYPTION CONTEXT flag](#structures.md#signed-encryptio
 
 ### OnDecrypt
 
-This interface takes [decryption materials](#structures.md#decryption-materials) and
-a list of [encrypted data keys](#structures.md#encrypted-data-key) as input and
-MAY modify it with the following behavior:
+The following inputs are REQUIRED:
+
+- [Algorithm Suite](#algorithm-suites.md)
+- [Encryption Context](#structures.md#encryption-context)
+- [Encrypted Data Keys](#structures.md#encrypted-data-keys)
+
+This interface MAY perform the following behavior:
 
 - [Decrypt data key](#decrypt-data-key)
 
-If this keyring attempted the above behavior, and succeeded, it MUST output the modified [decryption materials](#structures.md#decryption-materials).
+If this keyring attempted the above behavior, and succeeded, it MUST output the resulting [data key materials](#structures.md#data-key-materials). The [algorithm suite](#algorithm-suites.md) in the result must match the input [algorithm suite](#algorithm-suites.md).
 
-If the keyring did not attempt the above behavior,
-the keyring MUST output the [decryption materials](#structures.md#decryption-materials) unmodified.
+If the keyring did not attempt the above behavior, it MUST produce no output.
 
 #### Decrypt Data Key
 
-If the encryption materials do contain a plaintext data key, OnDecrypt MUST NOT decrypt a data key.
-If the [decryption materials](#structures.md#decryption-materials) do not include a plaintext data key,
-OnDecrypt MAY decrypt a data key.
-
-Decrypt Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Decrypt Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
 
 - [plaintext data key](#structures.md#plaintext-data-key)
 - [keyring trace](#structures.md#keyring-trace)
@@ -127,14 +136,13 @@ Decrypt Data Key MAY modify the following fields in the [encryption materials](#
 To perform this behavior, the keyring attempts to retrieve a plaintext data key from the input list
 of [encrypted data keys](#structures.md#encrypted-data-key).
 
-If the keyring is able to succesfully get at least one plaintext data key from any [encrypted data key](#structures.md#encrypted-data-key)
-and the [decryption materials](#structures.md#decryption-materials) still do not include a plaintext data key,
-it SHOULD set one resulting plaintext data key on the [decryption materials](#structures.md#decryption-materials).
+If the keyring is able to succesfully get at least one plaintext data key from any [encrypted data key](#structures.md#encrypted-data-key),
+it SHOULD provide the resulting plaintext data key in the [data key materials](#structures.md#data-key-materials).
 
 If the keyring is unable to get any plaintext data key using the input [encrypted data keys](#structures.md#encrypted-data-key)
-the keyring MUST NOT not update the [decryption materials](#structures.md#decryption-materials).
+the keyring MUST NOT produce output.
 
-If OnDecrypt updates the [decryption materials](#data-structure.md#decryption-materials) with a plaintext data key,
+If OnDecrypt produces a plaintext data key,
 the [keyring trace](#structures.md#keyring-trace) returned by OnDecrypt MUST include one trace
 associated with this keyring that has the [DECRYPTED DATA KEY](#structures.md#encrypted-data-key) flag.
 Note that this trace MAY include more than one flag, for example the [VERIFIED ENCRYPTION CONTEXT flag](#structures.md#verified-encryption-context).
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -293,7 +293,7 @@ If the call to [KMS Decrypt](#kms-decrypt) succeeds OnDecrypt MUST verify the fo
 - verify that the `KeyId` field has a value equal to the [encrypted data key's key provider info](##structures.md#key-provider-info).
 - verify that the `Plaintext` is of a length that fits the [algorithm suite](#algorithm-suites.md) given in the decryption materials.
 
-If any of the above are not true, OnDecrpyt MUST fail.
+If any of the above are not true, OnDecrypt MUST fail.
 
 If the response is successfully verified, OnDecrypt MUST do the following with the response:
 
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -151,13 +151,13 @@ OnEncrypt MUST take [encryption materials](#structures.md#encryption-materials) 
 If this keyring is a [discovery keyring](#is-discovery), OnEncrypt MUST return the input
 [encryption materials](#structures.md#encryption-materials) unmodified.
 
-If the input [encryption materials](#structures.md#encryption-materials) do not contain a plaintext data key
+If the input does not include a plaintext data key
 and this keyring does not have a [generator](#generator) defined,
 OnEncrypt MUST not modify the [encryption materials](#structures.md#encryption-materials.md)
 and MUST fail.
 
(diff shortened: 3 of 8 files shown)


framework/keyring-interface.md
```diff
@@ -36,37 +36,48 @@ in this document are to be interpreted as described in [RFC 2119](https://tools.
...
-Generate Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Generate Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
 
 - [plaintext data key](#structures.md#plaintext-data-key)
 - [keyring trace](#structures.md#keyring-trace)
 
 To perform this behavior, the keyring generates a [plaintext data key](#structures.md#plaintext-data-key)
-and sets the resulting plaintext data key on the [encryption materials](#structures.md#encryption-materials).
+and includes the resulting plaintext data key in the [data key materials](#structures.md#data-key-materials)' 
+[data key materials](#structures.md#data-key-materials).
```
RustanLeino: Unless the `'` is some kind of mark-up I don't understand, it seems "data key materials" is repeated.

robin-aws: Nope, just a regular ownership apostrophe on a plural noun. :) Will remove the accidental second copy.

WesleyRosenblum:
Do OnEncrypt and OnDecrypt need to return the same Data Key Materials? OnDecrypt wouldn't populate the Encrypted Data Keys in the response, it doesn't seem ideal to me for the response to contain null values on success.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
+
+- fit the specification for the [key derivation algorithm](#algorithm-suites.md#key-derivation-algorithm)
+  included in this decryption material's [algorithm suite](#algorithm-suite).
+- consist of cryptographically secure (pseudo-)random bits.
+- be kept secret.
+
+The plaintext data key SHOULD be stored as immutable data.
 
-The [algorithm suite](#algorithm-suites.md) to be used for [encryption](#encrypt.md).
+The plaintext data key SHOULD offer an interface to zero the plaintext data key
```
WesleyRosenblum: One thing that came up in the review of the Java Keyring was that the object comparison method (equals) for plaintext data key should not be vulnerable to timing attacks, ie the byte array should be compared in a time-safe way and not short circuit comparison when a byte differs. You could mention something about that here

framework/cmm-interface.md
```diff
@@ -85,17 +85,10 @@ it MUST return [encryption materials](#structures.md#encryption-materials) appro
 
 The encryption materials returned MUST include the following:
 
-- [Algorithm Suite](#algorithm-suite.md)
-    - If the encryption materials request contains an algorithm suite, the encryption materials returned SHOULD contain the same algorithm suite.
-- Plaintext Data Key
-- [Encrypted Data Keys](#structures.md#encrypted-data-keys)
-    - Every encrypted data key in this list MUST correspond to the above plaintext data key. 
+- [Data Key Materials](#structures.md#data-key-materials.md)
+    - If the encryption materials request contains an algorithm suite, the data key materials returned SHOULD contain the same algorithm suite.
```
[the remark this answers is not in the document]
robin-aws: I've heard the CMM interface is permitted to elect to use a different algorithm suite than the one requested. This was implicitly allowed in the stateful version, since there was no requirement that said it couldn't be mutated.

framework/keyring-interface.md
```diff
@@ -36,37 +36,48 @@ in this document are to be interpreted as described in [RFC 2119](https://tools.
...
 - [Generate data key](#generate-data-key)
 - [Encrypt data key](#encrypt-data-key)
 
 If this keyring attempted any of the above behaviors, and successfully completed those behaviors,
-it MUST output the modified [encryption materials](#structures.md#encryption-materials).
+it MUST output the result as [data key materials](#structures.md#data-key-materials).
 
-If the keyring did not attempt any of the above behaviors, it MUST output the
-[encryption materials](#structures.md#encryption-materials) unmodified.
+If the keyring did not attempt any of the above behaviors, it MUST produce no output.
```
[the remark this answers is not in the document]
robin-aws: "The interface MUST fail" is the equivalent language for throwing an exception. I'm using "produce no output" to mean produce null or None or something equivalent rather than a result or Some(something). We need something like this to describe interfaces that are allowed to "do nothing" A keyring can decide to not attempt to decrypt any encrypted data keys, but that's different than attempting and failing.

framework/raw-aes-keyring.md
```diff
@@ -151,27 +149,30 @@ the keyring MUST construct an [encrypted data key](#structures.md) with the foll
...
 
 ### On Decrypt
 
-On decrypt MUST take [decryption materials](#structures.md#decryption-materials) and
-a list of [encrypted data keys](#structures.md#encrypted-data-keys) as input.
+The following inputs are REQUIRED:
+
+- [Algorithm Suite](#algorithm-suites.md)
+- [Encryption Context](#structures.md#encryption-context)
+- [Encrypted Data Keys](#structures.md#encrypted-data-keys)
```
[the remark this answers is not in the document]
robin-aws: Agreed, there are a few cases with redundant clauses like this but I don't think its that helpful.

framework/kms-keyring.md
```diff
@@ -291,23 +293,22 @@ If the call to [KMS Decrypt](#kms-decrypt) succeeds OnDecrypt MUST verify the fo
...
   - The string field KeyName contains [generator](#generator).
   - The [flags](#structures.md$flags) field of this record includes exactly the following flags:
     - [DECRYPTED DATA KEY](#structures.md#supported-flags)
     - [VERIFIED ENCRYPTION CONTEXT](#structures.md#supported-flags)
-- immediately return the modified [decryption materials](#structures.md#decryption-materials).
+- immediately return the [data key materials](#structures.md#decryption-materials).
 
 If OnDecrypt fails to successfully decrypt any [encrypted data key](#structures.md#encrypted-data-key),
-then OnDecrypt MUST output the unmodified input [decryption materials](#structures.md#decryption-materials).
+then OnDecrypt MUST produce no output.
```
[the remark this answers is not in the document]
robin-aws: See previous reply :)

framework/structures.md
robin-aws: Good point. I cut #57 to track that.

framework/keyring-interface.md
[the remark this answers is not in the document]
WesleyRosenblum: What if it threw something like a EncryptionNotRequiredException? Just trying to see if we can avoid returning null. It could return Optional<DataKeyMaterials> but I'm not sure I love that either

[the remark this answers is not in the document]
robin-aws: Nah, not a fan of that since it's encouraging exceptions-as-control-flow. This isn't an exceptional situation so it shouldn't produce an exception. :)

Note that the abstract descriptions of behaviour can map to actual implementation in non-trivial ways. "produce no output" might be "don't populate any of the out parameters" if that is unambiguous (and that's what the older version was dictating in a more restrictive way).

robin-aws:
> Do OnEncrypt and OnDecrypt need to return the same Data Key Materials? OnDecrypt wouldn't populate the Encrypted Data Keys in the response, it doesn't seem ideal to me for the response to contain null values on success.

I'm waffling on this myself - it feels more convenient than correct. :) I might prefer to specify "multiple return values" instead, with phrases like "the Foo interface MUST produce an A and a B" if there's no value in packaging them up into a concept that's nothing more than a result tuple. Many implementations would create a "FooResult" structure in practice to implement that, but some languages will support multiple return values natively and won't need it.

framework/raw-aes-keyring.md
```diff
@@ -122,19 +122,17 @@ The authentication tag outputted from the AES-GCM encryption.
...
+If the input does not contain a plaintext data key, OnEncrypt MUST generate a random plaintext data key.
 
-The keyring MUST encrypt the plaintext data key in the [encryption materials](#structures.md#encryption-materials)
-using AES-GCM.
+The keyring MUST encrypt the plaintext data key using AES-GCM.
 
 The keyring must use AES-GCM with the following specifics:
 
-- it uses the [encryption context in the encryption materials](#structures.md#encryption-materials)
+- it uses the specified [encryption context](#structures.md#encryption-context)
```
lavaleri: s/specified/input

framework/keyring-interface.md
```diff
@@ -36,37 +36,48 @@ in this document are to be interpreted as described in [RFC 2119](https://tools.
...
+
+- [Algorithm Suite](#algorithm-suites.md)
+- [Encryption Context](#encryption-context)
+
+The following inputs are OPTIONAL:
+
+- [Plaintext Data Key](#structures.md#plaintext-data-key)
+
+This interface MAY return [data key materials](#structures.md#data-key-materials) appropriate for the request. If it does return this output
+and a plaintext data key was provided as input, the output must specify the same plaintext data key.
```
mattsb42-aws: `on encrypt` MUST always emit data key materials. They MAY have additional contents that were not present in the input materials.


Changing to "return nothing means nothing changed" is a fundamental change to the keyring interface and seems to place the onus of consolidating materials on the caller rather than the keyring.

framework/raw-aes-keyring.md
```diff
@@ -196,19 +197,19 @@ If decrypting, the keyring MUST use AES-GCM with the following specifics:
...
 If a decryption succeeds, this keyring MUST:
 
 - append a new [record](#structures.md#record) to the [keyring trace](#structures.md#keyring-trace)
-  in the input [decryption materials](#structures.md#decryption-materials).
+  in the output [data key materials](#structures.md#data-key-materials).
   This record MUST contain this keyring's [key name](#key-name) and [key namespace](#key-namespace),
   and the [flags](#structures.md$flags) field of this record MUST include the following flags:
   - [DECRYPTED DATA KEY](#structures.md#supported-flags)
-- add the resulting plaintext data key to the decryption materials and return the modified materials.
+- specify the resulting plaintext data key in the output [data key materials](#structures.md#data-key-materials).
```
lavaleri: s/specify/set

framework/keyring-interface.md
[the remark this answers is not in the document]
mattsb42-aws: > Nah, not a fan of that since it's encouraging exceptions-as-control-flow. This isn't an exceptional situation so it shouldn't produce an exception. :)

This is a language-specific design choice and as such should not be enshrined in the specification. This is why we used the "must fail" terminology; it leaves the decision to the implementor how specifically to manifest "failure" in their implementation.

> Note that the abstract descriptions of behaviour can map to actual implementation in non-trivial ways. "produce no output" might be "don't populate any of the out parameters" if that is unambiguous (and that's what the older version was dictating in a more restrictive way).

We should not be intentionally opening the spec to interpretation in this way. "How do I fail" is a very different thing than "maybe I'll emit something, maybe I won't, maybe that means something, maybe it doesn't".

framework/keyring-interface.md
mattsb42-aws: "includes the data key in the data key materials' data key materials"???

I have no idea what that is supposed to mean...do the data key materials separately contain data key materials?? At best that is confusing.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
-plaintext data key from this set of encryption materials.
+plaintext data key from this set of data key materials.
 
 If the plaintext data key is not included on this set of encryption materials, this list MUST be empty.
 
-##### Encryption Context
+##### Keyring Trace
 
-The [encryption context](#encryption-context) associated with this [encryption](#encrypt.md).
+An optional [keyring trace](#keyring-trace) containing all of the actions that keyrings have taken on this set of data key materials.
```
lavaleri: This gets a bit wonky when thinking about what things like multikeyrings should return. Technically the multikeyring is the only one interacting with the set of data materials it returns, and it doesnt directly encrypt/decrypt anything, so should it return an empty keyring trace? We should more strongly tie the keyring trace with the plaintext data key and encrypted data keys.

Maybe better wording would be more along the lines of: "all the actions that keyrings have taken related to the pdk and edks in this set of data key materials"

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
 #### Implementations
 
-- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/model/EncryptionMaterials.java)
-- [Python](https://github.com/aws/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/materials_managers/__init__.py)
-- [C](https://github.com/aws/aws-encryption-sdk-c/blob/master/include/aws/cryptosdk/materials.h)
-- [Javascript](https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts)
+- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/DataKey.java)
 
-#### Structure
+TODO: The above is only a similar data structure related to the deprecated master key/master key provider concepts.
```
mattsb42-aws: then don't list it as an implementation

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
-##### Algorithm Suite
+##### Plaintext Data Key
+
+A data key to be used as input for [encryption](#encrypt.md).
+
+The plaintext data key MUST:
+
+- fit the specification for the [key derivation algorithm](#algorithm-suites.md#key-derivation-algorithm)
+  included in this decryption material's [algorithm suite](#algorithm-suite).
+- consist of cryptographically secure (pseudo-)random bits.
```
mattsb42-aws: I specifically remember a lengthy discussion around wording regarding the entropy of the plaintext data key. There should be a definition elsewhere in this spec.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
+- be kept secret.
```
mattsb42-aws: As an implementor of this specification, what does this mean?
If we cannot clearly and concisely define this, it should not be a requirement.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
 If the plaintext data key is not included on this set of encryption materials, this list MUST be empty.
 
-##### Encryption Context
+##### Keyring Trace
 
-The [encryption context](#encryption-context) associated with this [encryption](#encrypt.md).
+An optional [keyring trace](#keyring-trace) containing all of the actions that keyrings have taken on this set of data key materials.
 
-##### Keyring Trace
+### Encryption Materials
```
lavaleri: The lack of symmetry between encryption materials and decryption materials feels a bit off, especially since the decrypt materials path needs to make the translation from data key materials to decryption materials, while the get encryption materials path uses the data key materials directly.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
+A data key to be used as input for [encryption](#encrypt.md).
+
+The plaintext data key MUST:
+
+- fit the specification for the [key derivation algorithm](#algorithm-suites.md#key-derivation-algorithm)
+  included in this decryption material's [algorithm suite](#algorithm-suite).
+- consist of cryptographically secure (pseudo-)random bits.
+- be kept secret.
+
+The plaintext data key SHOULD be stored as immutable data.
```
mattsb42-aws: Why is the data structure being immutable important here? What security properties does that add?

I would argue that we don't really care about the mutability of the data structure; we care about the ability for the language/runtime/os/etc to manage and destroy the contents of that data structure.

Again, I specifically remember a lengthy discussion around the requirements for plaintext data keys...

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
 
-The [algorithm suite](#algorithm-suites.md) to be used for [encryption](#encrypt.md).
+The plaintext data key SHOULD offer an interface to zero the plaintext data key
```
mattsb42-aws: The actor here should not be the data key, it should be the language/runtime/implementation/etc.

framework/structures.md
mattsb42-aws: The keyring trace is not optional.

framework/keyring-interface.md
```diff
@@ -75,66 +86,65 @@ perform the [Encrypt Data Key](#encrypt-data-key) behavior.
...
 
-If the [encryption materials](#structures.md#encryption-materials) contain a plaintext data key,
-OnEncrypt MAY encrypt a data key.
-If the encryption materials do not contain a plaintext data key, OnEncrypt MUST NOT encrypt a data key.
+If the input contains a plaintext data key, or if the keyring generates a data key,
+OnEncrypt MAY encrypt the data key.
+Otherwise, OnEncrypt MUST NOT encrypt a data key.
 
-Encrypt Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Encrypt Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
```
mattsb42-aws: "populate" implies that they are empty. What about keyrings that receive input that already contains EDKs? How about "add to"?

framework/keyring-interface.md
```diff
@@ -75,66 +86,65 @@ perform the [Encrypt Data Key](#encrypt-data-key) behavior.
...
 The [encrypted data keys](#structures.md#encrypted-data-key) produced by this keyring MUST
 have [ciphertexts](#structures.md#ciphertext) that can be decrypted to the plaintext data key in the
-[encryption materials](#structures.md#encryption-materials).
+[data key materials](#structures.md#data-key-materials).
 
-If OnEncrypt updates the [encryption materials](#data-structure.md#encryption-materials) with at least
-one new [encrypted data key](#structures.md#encrypted-data-key),
+If OnEncrypt outputs [data key materials](#structures.md#data-key-materials) with at least
+one [encrypted data key](#structures.md#encrypted-data-key),
 the [keyring trace](#structures.md#keyring-trace) returned by OnEncrypt MUST include at least one trace
```
mattsb42-aws: This wording changes the intent, back to my point about return values and data key material composition.

framework/keyring-interface.md
```diff
@@ -75,66 +86,65 @@ perform the [Encrypt Data Key](#encrypt-data-key) behavior.
...
+- [Encryption Context](#structures.md#encryption-context)
+- [Encrypted Data Keys](#structures.md#encrypted-data-keys)
+
+This interface MAY perform the following behavior:
 
 - [Decrypt data key](#decrypt-data-key)
 
-If this keyring attempted the above behavior, and succeeded, it MUST output the modified [decryption materials](#structures.md#decryption-materials).
+If this keyring attempted the above behavior, and succeeded, it MUST output the resulting [data key materials](#structures.md#data-key-materials). The [algorithm suite](#algorithm-suites.md) in the result must match the input [algorithm suite](#algorithm-suites.md), and the [encrypted data keys](#structures.md#encrypted-data-keys)
+list must be empty.
```
mattsb42-aws: > and the encrypted data keys list must be empty

What? This sounds like you are saying that a keyring would modify the EDK list. That would be a HUGE breaking change and I haven't seen any justification for it...

framework/multi-keyring.md
```diff
@@ -72,42 +71,34 @@ does not include a plaintext data key, OnEncrypt MUST fail.
...
+the [data key materials](#structures.md#data-key-materials) formed by merging the results of all OnEncrypt calls.
 
 ### OnDecrypt
 
-If the input [decryption materials](#structures.md#decryption-materials) contains a plaintext data key,
-OnDecrypt MUST immediately return the unmodified decryption materials.
-
-Otherwise, OnDecrypt MUST attempt to decrypt the [encrypted data keys](#structures.md#encrypted-data-key)
-in the input [decryption materials](#structures.md#decryption-materials) using it's
+OnDecrypt MUST attempt to decrypt the inputted [encrypted data keys](#structures.md#encrypted-data-key) using its
```
mattsb42-aws: There is a lot of intent lost here. Why?

framework/keyring-interface.md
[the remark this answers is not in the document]
robin-aws: I'm trying to introduce "produce no output" as distinct behaviour from "fail". In functional style this would mean, for example, returning an `Option<T>`. "producing no output" would map to returning `None`, but "failing" would make to raising an exception. Or in some cases you would return an `Either<Option<T>, Error>`, in which case "producing no output" would mean returning `Left(None)` whereas "failing" would mean returning `Right(Error(...))`.

I completely agree that we shouldn't dictate how either behaviour is implemented, but they must be implemented as distinguishable behaviours somehow.

framework/structures.md
robin-aws: Yeah agreed - I'm going to describe how the existing implementations inline the structure instead.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
 
 #### Implementations
 
-- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/model/EncryptionMaterials.java)
-- [Python](https://github.com/aws/aws-encryption-sdk-python/blob/master/src/aws_encryption_sdk/materials_managers/__init__.py)
-- [C](https://github.com/aws/aws-encryption-sdk-c/blob/master/include/aws/cryptosdk/materials.h)
-- [Javascript](https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts)
+- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/DataKey.java)
 
-#### Structure
```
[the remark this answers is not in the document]
robin-aws: Not intentional, good catch.

framework/keyring-interface.md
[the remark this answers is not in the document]
robin-aws: And in that case I'd rather not return it at all from Keyring operations since that's simpler. :) I'm starting to lean more towards just returning multiple values rather than a structure.

framework/structures.md
```diff
@@ -118,70 +122,78 @@ The encryption context MUST reserve the following key fields for use by the AWS
...
+TODO: The above is only a similar data structure related to the deprecated master key/master key provider concepts.
 
-Encryption materials are a structure containing materials needed for [encryption](#encrypt.md).
-This structure MAY include any of the following fields:
+##### Algorithm Suite
```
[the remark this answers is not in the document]
robin-aws: Yeah I agree - I'm thinking more that I would rather express the required relationship between input and output values, rather than duplicating values just so the requirement is only on the output as a separate structure.

framework/keyring-interface.md
```diff
@@ -36,37 +36,48 @@ in this document are to be interpreted as described in [RFC 2119](https://tools.
...
 
-Generate Data Key MAY modify the following fields in the [encryption materials](#structures.md#encryption-materials):
+Generate Data Key MAY populate the following fields in the [data key materials](#structures.md#data-key-materials):
 
 - [plaintext data key](#structures.md#plaintext-data-key)
 - [keyring trace](#structures.md#keyring-trace)
 
 To perform this behavior, the keyring generates a [plaintext data key](#structures.md#plaintext-data-key)
-and sets the resulting plaintext data key on the [encryption materials](#structures.md#encryption-materials).
+and includes the resulting plaintext data key in the [data key materials](#structures.md#data-key-materials)'
```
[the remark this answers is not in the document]
robin-aws: Already fixed :)

framework/keyring-interface.md
robin-aws: Stupid typo, mea cupla. Fixed.

framework/keyring-interface.md
```diff
@@ -75,66 +86,65 @@ perform the [Encrypt Data Key](#encrypt-data-key) behavior.
...
+If this keyring attempted the above behavior, and succeeded, it MUST output the resulting [data key materials](#structures.md#data-key-materials). The [algorithm suite](#algorithm-suites.md) in the result must match the input [algorithm suite](#algorithm-suites.md), and the [encrypted data keys](#structures.md#encrypted-data-keys)
+list must be empty.
 
-If the keyring did not attempt the above behavior,
-the keyring MUST output the [decryption materials](#structures.md#decryption-materials) unmodified.
+If the keyring did not attempt the above behavior, it MUST produce no output.
 
 #### Decrypt Data Key
 
-If the encryption materials do contain a plaintext data key, OnDecrypt MUST NOT decrypt a data key.
```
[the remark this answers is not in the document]
robin-aws: Actually the error is that I didn't remove the plaintext data key from the decryption materials - that was an "output parameter" for `Keyring.OnDecrypt` so it should no longer be part of the input, which means this statement no longer makes sense.

framework/raw-aes-keyring.md
robin-aws: I think I need different language - "set" implies mutating an existing object too strongly, so I worry about confusing the reader. I'm not sure what would be better though. Perhaps "include"?

framework/structures.md
lavaleri: Is the language in https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/data-format/message-header.md#message-id what you are referring to?

framework/keyring-interface.md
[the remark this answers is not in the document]
lavaleri: Ok, so what that does is move the responsibility of "deciding" whether to actually do OnDecrypt (in the case that the caller already has a plaintextDataKey) on the caller instead of the keyring itself.

This should be allowed, and is what the C implementation already does, but I'm unsure how we want to capture this in the spec, or if leaving any description out entirely is the right answer.

framework/keyring-interface.md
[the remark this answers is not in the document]
lavaleri: That's where I'm leaning too, and in that case we'd need to ensure that we specify that the outputted plaintextDataKey meets the spec for the AlgorithmSuiteID on input, since it would no longer be tied to the plaintextDataKey through the DataKeyMaterials.

framework/raw-aes-keyring.md
lavaleri: I like includes :+1:

mattsb42-aws:
I realized this morning that I don't think I ever summarized my issues with this change.

Rather than simply modifying the promises around keyring side effects, by stating that "`OnEncrypt`/`OnDecrypt` only return something if they do something and if they return something it's pieces that the caller must assemble", this change proposes two things:

1. A fundamental alteration of the keyring interface.
1. A shift in the responsibility for assembling encryption materials components.

This is a significant change from the current statement that "`OnEncrypt` always emits encryption materials and `OnDecrypt` always emits decryption materials". In this paradigm, it is *always* the responsibility of the keyring to assemble whatever it contributes.

The change as proposed also removes the binding between the various components of the encryption/decryption materials.

If the main thing you are after is the removal of side effects, we can do that without fundamentally changing the interface and boundaries of responsibility. However, I think that making even that a MUST rather than a SHOULD requires significant justification because it might not always be viable for every language.

robin-aws:
@mattsb42-aws and I discussed this PR offline, and I'm planning to cut a separate issue along the lines of "describe implementations such that pure functional implementations are allowed". This will help describe the motivation and more clearly describe what the actual changes are (since the diff is just awful to try to understand by itself). We should be able to avoid describing things in terms of mutation, but I will likely want to be less minimalistic about describing inputs and outputs: there are advantages to keeping related data together in `Encryption/DecryptionMaterials` structures even if some of the data is redundant in the context of operation results.

robin-aws:
Separate issue cut on this topic: #66 Closing this and will revisit at a later date.

PR #62, opened 2019-12-18
Defining the Master Key Provider Keyring

WesleyRosenblum:
*Issue #, if available:*

*Description of changes:*

Initial definition of the Master Key Provider Keyring.

### DIFF
--- a/framework/master-key-provider-keyring.md
+++ b/framework/master-key-provider-keyring.md
@@ -9,7 +9,6 @@
 
 ## Implementations
 
-- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/keyrings/MasterKeyProviderKeyring.java)
 - [Python](https://github.com/aws/aws-encryption-sdk-python/src/aws_encryption_sdk/keyrings/master_key.py)
 
 ## Overview
--- a/framework/master-key-provider-keyring.md
+++ b/framework/master-key-provider-keyring.md
@@ -1,0 +1,125 @@
+[//]: # (Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.)
+[//]: # (SPDX-License-Identifier: CC-BY-SA-4.0)
+
+# Master Key Provider Keyring
+
+## Version
+
+0.1.0-preview
+
+## Implementations
+
+- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/keyrings/MasterKeyProviderKeyring.java)
+- [Python](https://github.com/aws/aws-encryption-sdk-python/src/aws_encryption_sdk/keyrings/master_key.py)
+
+## Overview
+
+A keyring which wraps a legacy [Master Key Provider](master-key-provider-interface.md) to facilitate transitioning to [Keyrings](./keyring-interface.md).
+
+### Legacy
+
+Master Key Providers are a legacy construct that have been supplanted by [Keyrings](./keyring-interface.md). 
+The Master Key Provider Keyring should only be used when interfacing with Master Key Providers that cannot otherwise be
+transitioned to keyrings.
+
+## Definitions
+
+### Conventions used in this document
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Inputs
+
+On keyring initialization, a keyring MUST define the following:
+
+- [Master Key Provider](#master-key-provider)
+
+### Master Key Provider
+
+A [master key provider](master-key-provider-interface.md) that can decrypt data keys using available [master keys](master-key-interface.md)
+ as well as provide [master keys](master-key-interface.md) for use in encryption.
+
+## Operation
+
+### OnEncrypt
+
+This keyring MUST acquire a list of [master keys](master-key-interface.md) by calling the master key provider's [Get Master Keys for Encryption](master-key-provider-interface.md#get-master-keys-for-encryption)
+using the [encryption context](structures.md#encryption-context) from the [encryption materials](structures.md#encryption-materials) as input. If the 
+master key provider does not provide any master keys, OnEncrypt MUST fail.
+
+If the input [encryption materials](structures.md#encryption-materials) do not contain a [plaintext data key](structures.md#plaintext-data-key), OnEncrypt MUST
+attempt to generate a new plaintext data key and encrypt that data key by calling [Generate Data Key](master-key-interface.md#generate-data-key)
+on the first (primary) [master key](master-key-interface.md) provided by the master key provider.
+
+If the call to Generate Data Key does not succeed, OnEncrypt MUST NOT modify the encryption materials and MUST fail.
+
+If the call succeeds, OnEncrypt MUST do the following with the response from [Generate Data Key](master-key-interface.md#generate-data-key)
+- set the plaintext data key on the [encryption materials](structures.md#encryption-materials)
+- append a new [encrypted data key](structures.md#encrypted-data-key) to the encrypted data key list
+  in the [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - the [ciphertext](structures.md#ciphertext) is the response encrypted data key.
+  - the [key provider id](structures.md#key-provider-id) is the provider id of the primary master key.
+  - the [key provider information](structures.md#key-provider-information) is the key id of the primary master key.
+- append a new record to the [keyring trace](structures.md#keyring-trace) 
+  in the input [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - The string field KeyNamespace contains the provider id of the primary master key.
+  - The string field KeyName contains the key id of the primary master key.
+  - The [flags](structures.md#flags) field of this record includes the following flags:
+    - [GENERATED DATA KEY](structures.md#flags)
+    - [ENCRYPTED DATA KEY](structures.md#flags)
+    - [SIGNED ENCRYPTION CONTEXT](structures.md#flags) (if the master key supports encryption context signing)
+
+Next, for each [master key](master-key-interface.md) provided by the master key provider's 
+[Get Master Keys for Encryption](master-key-provider-interface.md#get-master-keys-for-encryption), the keyring MUST call
+ [EncryptDataKey](master-key-interface.md#encrypt-data-key). If the primary master key was used to generate a data key,
+ that master key MUST NOT be used again for encryption.
+ 
+ The input to [EncryptDataKey](master-key-interface.md#encrypt-data-key) MUST contain the following from the [encryption materials](structures.md#encryption-materials):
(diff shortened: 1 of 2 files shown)


framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+0.1.0-preview
+
+## Implementations
+
+- [Java](https://github.com/aws/aws-encryption-sdk-java/blob/master/src/main/java/com/amazonaws/encryptionsdk/keyrings/MasterKeyProviderKeyring.java)
+- [Python](https://github.com/aws/aws-encryption-sdk-python/src/aws_encryption_sdk/keyrings/master_key.py)
+
+## Overview
+
+A keyring which wraps a legacy [Master Key Provider](master-key-provider-interface.md) to facilitate transitioning to [Keyrings](./keyring-interface.md).
```
mattsb42-aws:
```suggestion
A keyring that wraps a legacy [Master Key Provider](master-key-provider-interface.md) to facilitate transitioning to [Keyrings](./keyring-interface.md).
```

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+## Inputs
+
+On keyring initialization, a keyring MUST define the following:
+
+- [Master Key Provider](#master-key-provider)
+
+### Master Key Provider
+
+A [master key provider](master-key-provider-interface.md) that can decrypt data keys using available [master keys](master-key-interface.md)
+ as well as provide [master keys](master-key-interface.md) for use in encryption.
```
mattsb42-aws: I would leave the definition of a MKP to the MKP spec. Notably, this definition implies that the MKP provided MUST work on both encrypt and decrypt paths, which might not always be correct.

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Operation
+
+### OnEncrypt
+
+This keyring MUST acquire a list of [master keys](master-key-interface.md) by calling the master key provider's [Get Master Keys for Encryption](master-key-provider-interface.md#get-master-keys-for-encryption)
```
mattsb42-aws: [NP/Formatting] This is a pretty long line. I don't think we have a max-length rule for this repo yet but my general rule of thumb is semantic newlines and stay below ~90 characters. (/me makes a note to go looking for markdown linters again).

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+using the [encryption context](structures.md#encryption-context) from the [encryption materials](structures.md#encryption-materials) as input. If the 
+master key provider does not provide any master keys, OnEncrypt MUST fail.
+
+If the input [encryption materials](structures.md#encryption-materials) do not contain a [plaintext data key](structures.md#plaintext-data-key), OnEncrypt MUST
+attempt to generate a new plaintext data key and encrypt that data key by calling [Generate Data Key](master-key-interface.md#generate-data-key)
+on the first (primary) [master key](master-key-interface.md) provided by the master key provider.
```
mattsb42-aws: Again, leave this to the MKP spec.

From MKP#Get_Master_Keys_For_Encryption:
> The output of this interface MUST indicate which one of those master keys MUST be used to generate the data key.

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+- set the plaintext data key on the [encryption materials](structures.md#encryption-materials)
+- append a new [encrypted data key](structures.md#encrypted-data-key) to the encrypted data key list
+  in the [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - the [ciphertext](structures.md#ciphertext) is the response encrypted data key.
+  - the [key provider id](structures.md#key-provider-id) is the provider id of the primary master key.
+  - the [key provider information](structures.md#key-provider-information) is the key id of the primary master key.
+- append a new record to the [keyring trace](structures.md#keyring-trace) 
+  in the input [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - The string field KeyNamespace contains the provider id of the primary master key.
+  - The string field KeyName contains the key id of the primary master key.
```
mattsb42-aws: Types and specific names are Java-specific. Be generic.

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+
+If the call to Generate Data Key does not succeed, OnEncrypt MUST NOT modify the encryption materials and MUST fail.
+
+If the call succeeds, OnEncrypt MUST do the following with the response from [Generate Data Key](master-key-interface.md#generate-data-key)
+- set the plaintext data key on the [encryption materials](structures.md#encryption-materials)
+- append a new [encrypted data key](structures.md#encrypted-data-key) to the encrypted data key list
+  in the [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - the [ciphertext](structures.md#ciphertext) is the response encrypted data key.
+  - the [key provider id](structures.md#key-provider-id) is the provider id of the primary master key.
+  - the [key provider information](structures.md#key-provider-information) is the key id of the primary master key.
```
mattsb42-aws: We should add an exception pointer here that refers to the Raw AES Keyring for how `RawMasterKey` and `JCEMasterKey` differ from this.

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+- append a new record to the [keyring trace](structures.md#keyring-trace) 
+  in the input [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - The string field KeyNamespace contains the provider id of the primary master key.
+  - The string field KeyName contains the key id of the primary master key.
+  - The [flags](structures.md#flags) field of this record includes the following flags:
+    - [GENERATED DATA KEY](structures.md#flags)
+    - [ENCRYPTED DATA KEY](structures.md#flags)
+    - [SIGNED ENCRYPTION CONTEXT](structures.md#flags) (if the master key supports encryption context signing)
```
mattsb42-aws: How do I know if it does or does not?

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+
+Next, for each [master key](master-key-interface.md) provided by the master key provider's
```
mattsb42-aws:
```suggestion
For each remaining [master key](master-key-interface.md) provided by
```

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+  - The string field KeyName contains the key id of the primary master key.
+  - The [flags](structures.md#flags) field of this record includes the following flags:
+    - [GENERATED DATA KEY](structures.md#flags)
+    - [ENCRYPTED DATA KEY](structures.md#flags)
+    - [SIGNED ENCRYPTION CONTEXT](structures.md#flags) (if the master key supports encryption context signing)
+
+Next, for each [master key](master-key-interface.md) provided by the master key provider's 
+[Get Master Keys for Encryption](master-key-provider-interface.md#get-master-keys-for-encryption), the keyring MUST call
+ [EncryptDataKey](master-key-interface.md#encrypt-data-key). If the primary master key was used to generate a data key,
+ that master key MUST NOT be used again for encryption.
```
mattsb42-aws: I *think* this should be a SHOULD not a MUST. I'm open to being wrong though.

```suggestion
 that master key SHOULD NOT be used again for encryption.
```

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+    - [ENCRYPTED DATA KEY](structures.md#flags)
+    - [SIGNED ENCRYPTION CONTEXT](structures.md#flags) (if the master key supports encryption context signing)
+
+### OnDecrypt
+
+If the input [decryption materials](structures.md#decryption-materials) contain a plaintext data key,
+OnDecrypt MUST immediately return the unmodified decryption materials.
+
+Otherwise, OnDecrypt MUST attempt to decrypt the [encrypted data keys](structures.md#encrypted-data-key)
+in the input using [Decrypt Data Key](master-key-provider-interface.md#decrypt-data-key)
```
mattsb42-aws: They also MUST be attempted in order.

framework/master-key-provider-keyring.md
```diff
@@ -0,0 +1,125 @@
...
+
+If OnDecrypt fails to successfully decrypt the [encrypted data keys](structures.md#encrypted-data-key),
+then OnDecrypt MUST output the unmodified input [decryption materials](structures.md#decryption-materials).
+
+Otherwise, OnDecrypt MUST do the following with the response from [Decrypt Data Key](master-key-provider-interface.md#decrypt-data-key)
+- set the plaintext data key on the [decryption materials](structures.md#decryption-materials)
+- append a new record to the [keyring trace](structures.md#keyring-trace)
+  in the input [encryption materials](structures.md#encryption-materials), constructed as follows:
+  - The string field KeyNamespace contains the provider id of the master key that was used to decrypt the data key.
+  - The string field KeyName contains the key id of the master key that was used to decrypt the data key.
```
mattsb42-aws: Ditto type and names being Java-specific.

framework/master-key-provider-keyring.md
WesleyRosenblum: Looking at the other documents in the spec, it looks like ~120 characters was generally the limit being followed, which I think is still reasonable. This line is over that even, I'll break it up

framework/master-key-provider-keyring.md
WesleyRosenblum: Ah ok, the Java ESDK doesn't do that

framework/master-key-provider-keyring.md
WesleyRosenblum: I would prefer to default to the more strict MUST NOT, I can't think of why we would want the master key to be used twice and SHOULD NOT leaves the door open for that.

framework/master-key-provider-keyring.md
WesleyRosenblum: I don't think there is any specified way to determine that, other than knowing that KMS and Raw AES support it. I could change it to say (if the master key is known to support encryption context signing, such as the KMS Master Key and Raw AES Master Key)

framework/master-key-provider-keyring.md
WesleyRosenblum: I've removed the definition

framework/master-key-provider-keyring.md
WesleyRosenblum: I've specified now that the input to DecryptDataKey must contain the Encrypted Data Keys in their original order

framework/master-key-provider-keyring.md
WesleyRosenblum: Updated

framework/master-key-provider-keyring.md
WesleyRosenblum: Now it says: the key provider information is the key id of the master key that generated the key, with the exception of the raw AES master key. For a detailed description of this exception, see the [Raw AES Keyring specification]

framework/master-key-provider-keyring.md
WesleyRosenblum: Updated

WesleyRosenblum:
The Java ESDK cannot use the MasterKeyProviderKeyring internally due to backward compatibility issues, leaving only the Python ESDK as the potential implementation of this spec. Since there is only one language that would implement this, it doesn't need to be in the spec anymore so I'm closing this.

Issue #68, opened 2020-02-11
KeyIds definition is contradictory


juneb:
The definition of key IDs in a KMS keyring is contraditory:
https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md#key-ids

* "Key IDs is a list of strings identifying KMS CMKs, in ARN format." This is true only for C.
* A CMK alias (e.g. "alias/MyCryptoKey")  <<< not an ARN
* A key id, which is valid in some languages (not C), is missing.
* It does not use the correct KMS key ID format names (e.g. "alias name").
https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-id
* It implies that only ARNs must be well-formed. Why would we ever accept anything that isn't well-formed?

The definition of Generator has a similar issue:
* The `The string MUST be one of the following:` list omits key ID, which is valid on encrypt in all languages except C.

acioc:
PR merged. Tracking implementation in https://github.com/awslabs/aws-encryption-sdk-specification/issues/129

PR #69, opened 2020-02-11
Resolve issue #68

juneb:
https://github.com/awslabs/aws-encryption-sdk-specification/issues/68

Question: Should we define ARN as "Amazon Resource Name" or is that more confusing?

* Resolve contradiction: not always an ARN
* Remove well-formed. We would never accept anything that's not well-formed. This also eliminates the typo. 
* Simplify the explanation of why we need key ARNs in the decryption keyring.


seebees:
This was resolved by #123

PR #71, opened 2020-02-24
Specify optional algorithm suite on Encrypt input and clarify GetEncr…

lavaleri:
…yptionMaterials call

*Issue #, if available:*

*Description of changes:* Encrypt does not specify algorithmSuiteID as an optional input. This should be specified and clarified that if supplied this is used as input for GetEncryptionMaterials.

Also fixup to say that keyringTrace MUST be included in response.


client-apis/encrypt.md
```diff
@@ -53,6 +54,10 @@ A CMM that implements the [CMM interface](#cmm-interface.md).
 
 A Keyring that implements the [keyring interface](#keyring-interface.md).
 
+### Algorithm Suite
+
+The [algorithm suite](#algorithm-suite.md) that SHOULD be used for encryption.
```
MatthewBennington: If I understand correctly. The input Algorithm Suite has no direct relation to the algorithm suite used for Encryption (or signing). This like should probably read: 
```
The [algorithm suite](#algorithm-suite.md) that MUST be passed to the CMM.
```

or something along those lines.

lavaleri: It's hard to know where to draw this line, but I'm trying to define this for the user's standpoint. As far as someone interacting with the Encrypt API is concerned, unless the CMM being used thinks it knows better (remember that CMMs SHOULD return encryption materials with the same algorithmSuiteID as was in the GetEncryptionMaterials call), the algorithm suite ID included as part of the encrypt call SHOULD just be the algorithm suite that is used for the encryption process.

Someone can read further into the spec to see exactly how this input is used such that the concept defined above is true.

MatthewBennington: Sounds good to me!

client-apis/encrypt.md
```diff
@@ -67,24 +72,31 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
...
-
-This behavior MAY output the following:
-
 - [keyring trace](#structures.md#keyring-trace)
 
 To construct the outputs, some fields MUST be constructed using information obtained
 from a set of valid [encryption materials](#structures.md#encryption-materials).
 This behavior MUST obtain this set of [encryption materials](#structures.md#encryption-materials)
 by calling [Get Encryption Materials](#cmm-interface.md#get-encryption-materials) on a [CMM](#cmm-interface.md).
+
```
seebees: Why whitespace ?

client-apis/encrypt.md
```diff
@@ -67,24 +72,31 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
...
 If instead the user supplied a [keyring](#keyring-interface.md), this behavior MUST use a [default CMM](#default-cmm.md),
 constructed using the user supplied keyring as input.
-The call to [Get Encryption Materials](#cmm-interface.md#get-encryption-materials) MUST include the
-input [encryption context](#encryption-context), if supplied.
-If the length is known on the input [plaintext](#plaintext), this call MUST also include that value
-
-The [algorithm suite](#algorithm-suites.md) used in all aspects of this behavior MUST be the algorithm suite in the
+The call to [Get Encryption Materials](#cmm-interface.md#get-encryption-materials) on that CMM
+MUST be constructed as follows:
+
```
seebees: This seems like a lot of whitespace?

client-apis/encrypt.md
```diff
@@ -67,24 +72,31 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
...
+- Encryption Context: If provided, this is the [input encryption context](#encryption-context).
+  Otherwise, this is an empty encryption context.
+- Algorithm Suite: If provided, this is the [input algorithm suite](#algorithm-suite).
+  Otherwise, this field is not included.
+- Plaintext Length: If provided, this is the [input plaintext length](#plaintext-length).
+  Otherwise, if the length is known on the input [plaintext](#plaintext), this value MUST equal to
```
seebees: What happens if you know the length _and_ the provided value is smaller?

client-apis/encrypt.md
```diff
@@ -67,24 +72,31 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
 This behavior MUST output the following if the behavior is successful:
 
 - [message](#message.md)
-
-This behavior MAY output the following:
```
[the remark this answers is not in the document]
lavaleri: This was originally included as MAY in order to "support" MKPs/MKs. However, in practice we've found the language confusing. Because this specification is scoped to describing encryption strictly using keyrings, I fixed this up to make it clear that we should always be outputting the keyring trace.

Not necessarily in scope for this PR, so if there is any dissent here I'm happy to practice good PR hygiene and get this done in a separate PR.

client-apis/encrypt.md
lavaleri: I personally found it more readable. One section says "you MUST get encryption materials from a CMM", another says "this is how you get those materials".

client-apis/encrypt.md
lavaleri: Newline before a list is standard throughout the spec. It shouldn't affect how it's rendered on github, but some things (like pandocs) end up requiring it to properly render the list.

client-apis/encrypt.md
lavaleri: The API should honor the API contract first, so it should go with the smaller value, which means that it is the responsibility of the encrypt API to not encrypt more than that.

For the scope of this PR, this means that when getting encryption materials the provided value should be used, which should be what the language above says to do.

lavaleri:
For `maxBodySize` I will have a follow up PR.

client-apis/encrypt.md
seebees: I disagree.
I think...

Say we try and encrypt the string `secret`
and pass a length of `3`.
It is unacceptable to return an encrypted payload that will decrypt to `sec`.

I say we must error is we are unable to encrypt the message.
In the ESDK-JS,
if you pass plaintext with a knowable length,
I will clobber whatever optional length with the known value
https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/encrypt-node/src/encrypt.ts#L31
In the case of an unknown length,
the plaintext length is the maximum
https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/encrypt-node/src/framed_encrypt_stream.ts#L73

lavaleri: After talking offline:

1. We're in agreement that the example you give above showcases a behavior we want to make sure our API MUST NOT do.
2. We agree that if the API will always know the length of the plaintext (ie exclusively a one shot API), then the API MUST NOT give the user the option to specify a plaintext length.
3. In the case that it is possible for the plaintext length to be unknown, then the API MUST have an OPTIONAL input to specify the plaintext length.
3.5. In this case, if during the Encrypt operation, it turns out the plaintext length *is* known, and the user specified a plaintext length less than the actual plaintext length, Encrypt MUST NOT do the behavior described above.

So open question: What behavior MUST it do instead? As you said, JS clobbers the value. My preference would be to fail (in whatever way makes sense for the implementation).

Second question: What should be done if the supplied plaintext length is greater than the actual length? Should we return an error, should we use the actual length, or should we use the supplied length? All current implementations do the third, and I would argue that that allows users to depend on weird logic that is not the intended purpose of the plaintext length field.

Thoughts?

lavaleri: cutting specifying this from the scope of this PR. This will be addressed in #138.

Issue #73, opened 2020-03-11
AWS KMS keyring encrypt behavior is incorrect


mattsb42-aws:
## Intent

When a user configures a KMS keyring with key IDs, they are stating an intent that they need each of those CMKs to be able to independently decrypt the resulting encrypted message.

For example, if I configure a KMS keyring with CMKs A, B, and C and use that keyring to encrypt a message, then I expect three things to be true:

1. CMK A can decrypt the encrypted message.
1. CMK B can decrypt the encrypted message.
1. CMK C can decrypt the encrypted message.

If any of these are not true of the resulting encrypted message, then the keyring has failed to honor my intent.

The key IDs that I provide to the keyring are the CMKs that I want the keyring to use.

Separate from the statement of what CMKs I want the keyring to use, I can also provide a client supplier. The client supplier provides the KMS client that the keyring needs in order to interact with each CMK.

It is important to note, however, that the client supplier does NOT describe my intent for what CMKs I want to use. It instead describes my intent for how and where I want to talk to KMS.

Put another way, the key ID list is my expressed intent of *what* the KMS keyring needs to do while the client supplier's behavior is my expressed intent of *how* the KMS keyring needs to do that.

### Encrypt vs Decrypt

The intent of the AWS Encryption SDK framework on decrypt is very different than the intent on encrypt.

On encrypt, the caller describes their intent for the requirements to decrypt the resulting message.

On decrypt, the caller provides resources that attempt to do that decryption.

This is a fundamentally asymmetric relationship with very different implications. If we encounter a problem on encrypt, we cannot fully honor the decryption requirements and so we MUST stop. If we encounter a problem on decrypt, then within the bounds of the keyring responsibility all we know is that we have not *yet* located a resource that satisfies those decryption requirements.

### Requirements

This intent and behavior can be reduced to the following two requirements:

1. On encrypt, if any configured CMK cannot be used, that is an error and encryption MUST halt.
1. On decrypt, the keyring will never halt decryption because of a failure to decrypt.

## The Problem

In the KMS keyring specification, when discussing the generator CMK, we state[1]:

> If the [client supplier](#client-supplier) does not provide any client for the given region for this [KMS GenerateDataKey](#kms-generatedatakey) call, OnEncrypt MUST NOT modify the [encryption materials](#structures.md#encryption-materials.md)
and MUST fail.

In short: if the generator CMK fails for any reason then the `OnEncrypt` call MUST fail.

[...]

PR #74, opened 2020-03-16
docs: Address KMS keyring on-encrypt issues

mattsb42-aws:
*Issue #, if available:*
resolves: #73 

*Description of changes:*
This resolves the discrepancies found between the intent of KMS keyring behavior and configuration and the spec and adds a description of that intent.

### DIFF
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -212,7 +212,7 @@ For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted fro
 [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the client returned by the
 [client supplier](#client-supplier) when given that region as input.
 If an AWS region cannot be extracted from the Key ID then the KMS Keyring MUST input a value denoting an unknown region.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call, OnEncrypt MUST skip that particular CMK.
+If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call, OnEncrypt MUST fail.
 
 To encrypt the plaintext data key with a CMK, OnEncrypt MUST call [KMS Encrypt](#encrypt) with a request
 constructed as follows:
@@ -313,3 +313,9 @@ then OnDecrypt MUST output the unmodified input [decryption materials](#structur
 
 [TODO: What security properties are guaranteed by this keyring? Also, note how the security properties
 can vary drastically depending on key policies]
+
+## Contributing Issues
+
+This is a record of issues that contributed to this specification.
+
+* [#173 Resolve incorrect description of behavior of additional CMKs on encrypt.](https://github.com/awslabs/aws-encryption-sdk-specification/issues/73)
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -219,7 +219,10 @@ For each [KMS Encrypt](#kms-encrypt) call, if an AWS region can be extracted fro
 [KMS client](#kms-client) that calls [KMS Encrypt](#kms-encrypt) MUST be the client returned by the
 [client supplier](#client-supplier) when given that region as input.
 If an AWS region cannot be extracted from the Key ID then the KMS Keyring MUST input a value denoting an unknown region.
-If the [client supplier](#client-supplier) does not provide any client for the given region for this Encrypt call, OnEncrypt MUST fail.
+If the [client supplier](#client-supplier) does not provide any client
+for the given region for this [KMS Encrypt](#kms-encrypt) call,
+OnEncrypt MUST NOT modify the [encryption materials](#structures.md#encryption-materials.md)
+and MUST fail.
 
 To encrypt the plaintext data key with a CMK, OnEncrypt MUST call [KMS Encrypt](#encrypt) with a request
 constructed as follows:
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -349,7 +349,7 @@ When a user configures a KMS keyring for use on decrypt,
 they are stating their intent for which CMKs
 the keyring will *attempt* to use to decrypt encrypted data keys.
 
-For example, if a user configures a KMS keyring with CMK C
+For example, if a user configures a KMS keyring with CMK C (using the CMK ARN)
 and uses it to decrypt an encrypted message
 that contains encrypted data keys for CMKs A, B, and C,
 then the keyring will attempt to decrypt using CMK C.
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -73,6 +73,9 @@ and MAY return a KMS Client that can make the following API calls in the given A
 
 The keyring will use this client supplier to determine the KMS client to use when making KMS calls.
 
+The purpose of the client supplier is
+to control how the keyring communicates with AWS KMS.
+
 ### Key IDs
 
 Key IDs is a list of strings identifying KMS CMKs, in ARN format.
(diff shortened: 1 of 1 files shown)


framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+the keyring will *attempt* to use to decrypt encrypted data keys.
+
+For example, if a user configures a KMS keyring with CMK C
+and uses it to decrypt an encrypted message
+that contains encrypted data keys for CMKs A, B, and C,
+then the keyring will attempt to decrypt using CMK C.
+
+However, if the keyring attempts to decrypt using CMK C and cannot,
+this failure still honors the configured intent and MUST NOT halt decryption.
+The configured intent is that they keyring MUST *attempt* with these CMKs,
```
lavaleri: I would try to make it a bit more clear that `these CMKs` refers to the CMKs configured with this keyring, not "CMK A, B, and C". Maybe make this more clear by explicitly giving the example of "If you configure the KMS keyring for decrypt with CMKs C, D, E" or something like that? Or is that more confusing?

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+will halt message decryption.
+(See [Default Cryptographic Materials Manager](./default-cmm.md))
+
+### Requirements
+
+These goals can be reduced to the following two requirements:
+
+1. On encrypt, if any configured CMK cannot be used,
+    that is an error and encryption MUST halt.
+1. On decrypt, the keyring MUST never halt decryption because of a failure to decrypt.
```
lavaleri: NP: "the keyring MUST NOT halt decryption because of..."

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+It is instead attempting to satisfy the requirements that were set on encryption.
+If the keyring cannot satisfy those requirements it MUST NOT halt message decryption.
+
+No keyring can know if it is the last keyring to attempt decryption.
+If all keyrings are exhausted and none of them were able to decrypt an encrypted data key
+then the cryptographic materials manager that managed those keyrings
+will halt message decryption.
+(See [Default Cryptographic Materials Manager](./default-cmm.md))
+
+### Requirements
```
lavaleri: I like being able to state high level intentions like this, but I think it might be a bit confusing or miss-able underneath all of this soft discussion of intention. These requirements are invariants which the above specification should cover. i.e. If your implementation follows the above specification for behavior of this keyring, it will have these properties. These invariants are stated elsewhere to clarify intentions and highlight important properties that the above spec offers. I'm not quite sure how to get this across, but "Requirements" doesn't seem like the right title.

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+1. On encrypt, if any configured CMK cannot be used,
+    that is an error and encryption MUST halt.
+1. On decrypt, the keyring MUST never halt decryption because of a failure to decrypt.
+
 ## Security Considerations
 
 [TODO: What security properties are guaranteed by this keyring? Also, note how the security properties
 can vary drastically depending on key policies]
+
+## Contributing Issues
```
lavaleri: I like this idea, but it will be easy to forget to follow up on unless we're all on board. Maybe we could update the PR template to specifically call out to include such updates in the spec. (Doesn't have to be done as part of this PR, just a thought)

framework/kms-keyring.md
WesleyRosenblum:
```suggestion
The configured intent is that the keyring MUST *attempt* with these CMKs,
```

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+this failure still honors the configured intent and MUST NOT halt decryption.
+The configured intent is that they keyring MUST *attempt* with these CMKs,
+not that they MUST *succeed*.
+
+### Why OnEncrypt and OnDecrypt are different
+
+On encrypt, the user describes their intent
+for the requirements to decrypt the resulting message.
+Because of this,
+the keyring MUST create encryption materials that satisfies those requirements.
```
WesleyRosenblum:
```suggestion
the keyring MUST create encryption materials that satisfy those requirements.
```

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+it cannot fully honor the decryption requirements and so MUST halt message encryption.
+
+However, on decrypt the keyring is not creating anything.
+It is instead attempting to satisfy the requirements that were set on encryption.
+If the keyring cannot satisfy those requirements it MUST NOT halt message decryption.
+
+No keyring can know if it is the last keyring to attempt decryption.
+If all keyrings are exhausted and none of them were able to decrypt an encrypted data key
+then the cryptographic materials manager that managed those keyrings
+will halt message decryption.
```
WesleyRosenblum: I wonder if some of this belongs in the keyring interface spec, since it isn't specific to the KMS keyring

framework/kms-keyring.md
```diff
@@ -73,6 +73,9 @@ and MAY return a KMS Client that can make the following API calls in the given A
 
 The keyring will use this client supplier to determine the KMS client to use when making KMS calls.
 
+The purpose of the client supplier is
```
seebees: What about something like:

To communicate with AWS KMS the AWS KMS Keyring needs a client.
The client suppliers purpose is to configure that client.
An example might be different timeouts per region.
It is not intended to provide control flow or filtering.

I also wonder if we should lead with this?

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+and uses it to decrypt an encrypted message
+that contains encrypted data keys for CMKs A, B, and C,
+then the keyring will attempt to decrypt using CMK C.
+
+However, if the keyring attempts to decrypt using CMK C and cannot,
+this failure still honors the configured intent and MUST NOT halt decryption.
+The configured intent is that they keyring MUST *attempt* with these CMKs,
+not that they MUST *succeed*.
+
+### Why OnEncrypt and OnDecrypt are different
```
seebees: I know that this is in the KMS Keyring file.
But this wording starts to move towards Keyrings generally,
as opposed to this specific KMS Keyring.

framework/kms-keyring.md
seebees: I think we need to be careful here.
There are keyrings that I can envision that do not satisfy this "Any subcomponent can decrypt".
Most notably an N of M style keyring.

framework/kms-keyring.md
seebees: I think I agree. Perhaps examples? I know that clarity can be a form of repetition, but I think that the encrypt requirement is special here, and the decrypt requirement just flows from the keyring intent.

I also wonder if some of this intent language should not also be moved to the Multi-Keyring as these two share a lot of intent...

framework/kms-keyring.md
```diff
@@ -73,6 +73,9 @@ and MAY return a KMS Client that can make the following API calls in the given A
 
 The keyring will use this client supplier to determine the KMS client to use when making KMS calls.
 
+The purpose of the client supplier is
+to control how the keyring communicates with AWS KMS.
```
MatthewBennington:
```suggestion
to control how the KMS Keyring communicates with AWS KMS.
```

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+## Configuration Intent
+
+### OnEncrypt Goal
+
+When a user configures a KMS keyring with key IDs
+and use that keyring to encrypt a message,
+they are stating their intent that they need each one of those CMKs to be able to
+independently decrypt the resulting encrypted message.
+
+For example, if a user configures a KMS keyring with CMKs A, B, and C
```
MatthewBennington: IMO There should be a separate "Example Behaviors" section.

MatthewBennington:
```suggestion
For example, if a user configures a KMS Keyring with CMKs A, B, and C
```

framework/kms-keyring.md
```diff
@@ -309,7 +312,84 @@ If the response is successfully verified, OnDecrypt MUST do the following with t
...
+### OnDecrypt Goal
+
+When a user configures a KMS keyring for use on decrypt,
+they are stating their intent for which CMKs
+the keyring will *attempt* to use to decrypt encrypted data keys.
+
+For example, if a user configures a KMS keyring with CMK C
+and uses it to decrypt an encrypted message
+that contains encrypted data keys for CMKs A, B, and C,
+then the keyring will attempt to decrypt using CMK C.
```
MatthewBennington: This is only true if the string identifying the CMK is a well-formed key ARN, or a well-formed alias.

framework/kms-keyring.md
mattsb42-aws: I'm inclined to say that yes, some form of this should be in the interface spec. Maybe not these exact words, though. I don't think that this PR is the correct place to introduce that, though. I've created #75 to follow up on this.

In the case of either proposed n-of-m keyring, I think these statements still hold true. It is simply the n-of-m keyring that would be making the halt decision rather than one of the child keyrings.

[1] most likely; pending design work

framework/kms-keyring.md
mattsb42-aws:
```suggestion
1. On decrypt, the keyring MUST NOT halt decryption because of a failure to decrypt.
```

framework/kms-keyring.md
mattsb42-aws: > If your implementation follows the above specification for behavior of this keyring, it will have these properties

Agreed, but I see this as almost a parity check for that: this is a very concise, straight-forward, statement of the top-level requirements. `Operation` is a description of *how* to achieve those requirements and `Configuration Intent` is an explanation of *why* we have those requirements.

I agree that there might be a better place to put this. I'm not entirely sure where in the doc that is, but I'll play with some options. For a starting point, I wanted to make sure it was in there at all. :)

> I also wonder if some of this intent language should not also be moved to the Multi-Keyring as these two share a lot of intent...

Not moved, but replicated. I've created #76 to track that.

framework/kms-keyring.md
mattsb42-aws: Agreed. I created #77 to track this.

framework/kms-keyring.md
mattsb42-aws: Agreed. See #75

framework/kms-keyring.md
mattsb42-aws: Per guidance from @juneb, "keyring" is not a proper noun, so should not be capitalized.
```suggestion
to control how the KMS keyring communicates with AWS KMS.
```

framework/kms-keyring.md
mattsb42-aws: I feel that would be distracting in this particular case. What I'm trying to do here is demonstrate how this intent manifests as a "show by example" means of explaining the behavior.

framework/kms-keyring.md
seebees: I agree that on encrypt describes an intent that on decrypt will satisfy.

I'm saying that in your example of CMKs A, B, C; if I on Encrypt 2 of 3. Then it is not true that A, B, and C call all decrypt the message.  One of them will fail.

This is _not_ the intent for the AWS KMS Keyring, but other keyrings may have such an intent. Just wanting to be clear on how we describe this relationship.
I'll resolve this here, and I think that it is best moved over to #75.

mattsb42-aws: I think that will depend on the design of the n-of-m-decrypt keyring. What you described is not what I have in mind for the way such a thing would work, but I agree that this is not the place to design that.

Issue #98, opened 2020-04-28
Define how the default CMM uses master key providers


mattsb42-aws:
As a legacy-record effort, we need to define what the default CMM does with master key providers. This is particularly of import for customers migrating from master key providers to keyrings.

PR #100, opened 2020-05-01
docs: define how the default CMM uses a master key provider

mattsb42-aws:
*Issue #, if available:*
resolves: #98 

*Description of changes:*

This addition defines how the default CMM uses a master key provider.


framework/default-cmm.md
```diff
@@ -55,7 +59,10 @@ If the algorithm suite contains a [signing algorithm](algorithm-suites.md#signat
   - The key MUST be the reserved name, `aws-crypto-public-key`.
   - The value MUST be the base64-encoded public verification key.
 
-On each call to Get Encryption Materials, the default CMM MUST make a call to the key provider's [On Encrypt](keyring-interface.md#onencrypt) function.
+On each call to Get Encryption Materials,
+the default CMM MUST make a call to the keyring's [On Encrypt](keyring-interface.md#onencrypt) function.
+If the default CMM is configured with a master key provider instead,
```
seebees, line 64: I would make it clear here that master key provider is both legacy
and not implemented in every runtime here as well.

Or, we could leave this out
and have the description of mater key providers
be the only description.
In that case, I would suggest a callout in the Overview.

framework/default-cmm.md
```diff
@@ -69,10 +76,53 @@ The values obtained above MUST be included in the encryption materials returned.
 If the algorithm suite contains a [signing algorithm](algorithm-suites.md#signature-algorithm),
 the default CMM MUST remove the verification key from the encryption context.
 
-On each call to Decrypt Materials, the default CMM MUST make a call to the key provider's [On Decrypt](keyring-interface.md#ondecrypt) function.
+On each call to Decrypt Materials,
+the default CMM MUST make a call to the keyring's [On Decrypt](keyring-interface.md#ondecrypt) function.
+If the default CMM is configured with a master key provider instead,
+see [Decrypt Materials (master key provider)](#decrypt-materials-master-key-provider).
+
+The default CMM MUST obtain the following from the response:
```
seebees, line 84: Duplicate?

framework/default-cmm.md
lavaleri: +1
I think I'd prefer the latter, but at the very least if we mention MPKs in the main spec let's make it more clear that this is referring to legacy behavior that one SHOULD NOT implement in new implementations.

framework/default-cmm.md
mattsb42-aws: yup

framework/default-cmm.md
mattsb42-aws: Yeah, I think I'm ok with that. If someone knows they're looking for MKP behavior, they shouldn't have any issues finding it in the "Legacy" section, and the descriptions in that section make it clear where they fit into the spec.

Issue #87, opened 2020-03-30
Clarify halting vs non-halting failure behaviors


mattsb42-aws:
We've gone back and forth a lot on how to describe different kinds of failures, and a result of this is that some of these points in the spec are a bit opaque.

We should go through and clarify when we mean different kinds of failure and what they mean. I would propose that we separate these into two failures: failures that MUST halt the encrypt or decrypt operation and failures that MUST NOT halt the operation.

Issue #76, opened 2020-03-17
add statement of intent and top-level requirements for multi-keyring encrypt HALT


mattsb42-aws:
As discussed here[1], we should add similar explicit statements of intent and required behavior for the multi-keyring.

[1] https://github.com/awslabs/aws-encryption-sdk-specification/pull/74#discussion_r393796843

Issue #75, opened 2020-03-17
add statements of intent for keyring encrypt and decrypt behavior


mattsb42-aws:
As discussed here[1], we should add a statement of intent to the keyring interface spec that describes how and when a keyring MUST or MUST NOT halt operation.

[1] https://github.com/awslabs/aws-encryption-sdk-specification/pull/74#discussion_r393376352

Issue #11, opened 2019-08-20
Define how the specification should deal with errors/failures


lavaleri:
Throughout certain parts of the spec, we prescribe how successes should behave, but not failures/errors, or what failing to meet the requirements of the behavior means. (For example, when defining the keyring OnDecrypt behavior, we do not state what should happen when decryption is attempted but failed)

While it might be ok for the spec to be quiet about failure behaviors, this should be a deliberate choice. We need to identify areas of the spec which are inconsistent/ambigous regarding error/failure cases, and we should make a decision as to whether we definitely want to specify something about that behavior, or whether we want to leave these parts open to implementation, and the consequences of each.

Acceptance criteria:
- team alignment on spec direction regarding error cases
- new issues created to address problem spots in spec.

mattsb42-aws:
This also involves:

* #87
* #76
* #75

PR #131, opened 2020-06-05
docs: Add change doc introducing the concept of the keyring decryption contract

mattsb42-aws:
_Issue #, if available:_

This affects several issues, but does not resolve any.

#11 #75 #76 #87 

_Description of changes:_

This introduces a change doc that presents the concept of the keyring decryption contract.

I have not yet modified the keyring interface specification document. Once we have agreement on the change doc, I will proceed to that.

_Reviews:_

As always, any reviewer is welcome to provide any feedback that they think will add value, but I am looking for specific areas of feedback from the following individuals:

* @SalusaSecondus and @agray256 : Feedback on the decryption contract definition, framing, and how it can feed into defining the security properties of a keyring. Also for a perspective check if this is a concept that already exists with a different name that we should reference.


changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+
+Clearly defining what causes keyrings to fail
+makes it simpler for users to predict failing behavior
+based on either environment
+and across implementations.
+
+## Guide-level Explanation
+
+When a keyring succeeds on encryption,
+it creates an artifact: one or more encrypted data keys.
```
seebees, line 117:
```suggestion
On encryption a keyring
creates an artifact: one or more encrypted data keys.
```

I don't know that we care about the succeed/fail problem here.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+This artifact has certain requirements that
+MUST be met on decryption
+in order to obtain the plaintext data key.
+This set of requirements is what we call
+the "keyring decryption contract."
+The process that a keyring uses to create this artifact
+is largely irrelevant to the decryption contract.
```
seebees, line 124:
```suggestion
is only relevant as a mechanism to fulfill the decryption contract.
```

Or something positive.  The keyring is responsible for fulfilling the contract, and if the artifact can help enforce that great, otherwise, as you say it is not relevant.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+## Reference-level Explanation
+
+### Definitions and Terms
+
+- **writing the decryption contract** :
+  This refers to the process of taking the plaintext data key
+  and successfully creating an artifact
+  (one or more encrypted data keys)
+  that requires a decrypting keyring to
+  fulfill the encrypting keyring's decryption contract.
```
seebees, line 137:
```suggestion
  that a decrypting keyring can use to
  fulfill the keyring decryption contract.
```

The method of communication between keyrings is the artifact.
And I don't think we can really enforce needing a keyring to do this.

Once the contract is written is it immutable?
If so, then I think that this artifact _is_ the contract.

Still noodling.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+
+On encryption,
+if a keyring is unable or unwilling to
+write its decryption contract,
+that keyring MUST fail.
+
+On decryption,
+if a keyring is unable or unwilling to
+fulfill its decryption contract
+using the encrypted data keys that it receives,
```
seebees, line 153: Does the decryption keyring have a contract?  Or is it trying/able to fulfill the contracts it receives? e.g. the encrypted data keys?

mattsb42-aws: > Does the decryption keyring have a contract?

No, but it does need to be able to determine if it SHOULD be able to fulfill a given decryption contract (EDK).

changes/2020-06-04_how-to-fail-with-keyrings/change.md
mattsb42-aws: That's one that I could go either way on. On the one hand, stating that "this always happens on encryption" is inaccurate because if it fails, by definition that does not happen. On the other hand, I think you might be right that at this point in the doc we should just assume that everything is working correctly.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
seebees: Then what about this?

```suggestion
fulfill the decryption contract
of the encrypted data keys that it receives,
```

changes/2020-06-04_how-to-fail-with-keyrings/change.md
seebees: I think that the creation of the artifacts IS success.
At least in the context of this doc :)

changes/2020-06-04_how-to-fail-with-keyrings/change.md
robin-aws: +1, I think we should be consistent that "the decryption contract of a keyring" always refers to the decryption contract of the artifacts it produces when encrypting.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,223 @@
...
+do slightly different things,
+leading to implementation drift
+and confusion for users and implementors.
+
+In order to clarify this situation,
+we MUST first define what it means for a keyring to fail,
+and that means defining what it means for a keyring to succeed.
+To do this,
+we define a new concept of the
+"keyring decryption contract".
```
robin-aws, line 88: The "keyring" part here is still the main sticking point for me: is the decryption contract ultimately owned by the artifact, or is it attached to the keyring that produced it and the artifact inherits it?

I think this is the key question: if it is possible to encrypt using a Master Key and decrypt using a Keyring, does that mean the contract is ultimately a property of the artifact(s), and the "keyring decryption contract" should be "the decryption contract of the artifacts this keyring produces"? Otherwise we are stating that there is no contract involved in the migration path.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -108,37 +108,43 @@ the security properties of each keyring.
...
-and across implementations.
+in different environments and implementations.
 
 ## Guide-level Explanation
 
-When a keyring succeeds on encryption,
-it creates an artifact: one or more encrypted data keys.
+On encryption,
+a keyring creates an artifact:
+one or more encrypted data keys.
```
seebees, line 117: See #158. There is some cross pollination of words that might be good to have.
Not necessarily something to block on.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
mattsb42-aws: The line I'm trying to thread here is that:

1. These artifacts are created and consumed by keyrings.
1. The keyring's decryption contract MUST be fulfilled through the use of that artifact.
1. In future changes we SHOULD introduce a concept of a CMM decryption contract, which is different than the decryption contracts that are written and fulfilled by keyrings.

The keyring is the actor, so "keyring decryption contract". Unless we can come up with a better name, I would be inclined to say that that CMM analogue should be a "CMM decryption contract."

robin-aws: I see your point about disambiguating the scope of the contract. Would it be better to use "key decryption contract" instead?

changes/2020-06-04_how-to-fail-with-keyrings/background.md
```diff
@@ -0,0 +1,166 @@
...
+- We need to define detailed decryption contracts
+  (defined in [Issues and Alternatives](#issues-and-alternatives))
+  for each keyring defined in the ESDK specification.
+  This will be addressed separately once we have
+  consensus on this document.
+
+- A mechanism to interrogate a keyring to determine
+  the requirements of its decryption contract
+  (defined in [Issues and Alternatives](#issues-and-alternatives))
+  is out of scope for this document.
```
robin-aws, line 61: Link #144?

changes/2020-06-04_how-to-fail-with-keyrings/background.md
```diff
@@ -0,0 +1,166 @@
...
+
+  - The keyring MUST have access to the symmetric wrapping key.
+  - The encryption context MUST exactly match
+    the encryption context used on encrypt.
+
+    - NOTE: In compliant implementations this is a MUST,
+      but mathematically this is a SHOULD
+      because it is possible to create a keyring
+      that ignores this clause in the decryption contract
+      by using AES-CTR rather than AES-GCM.
```
SalusaSecondus, line 96: NitPick: I'd probably strike this aside as not necessary.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,223 @@
...
+is what we call the "keyring decryption contract."
+
+## Reference-level Explanation
+
+### Definitions and Terms
+
+- **decryption contract** :
+  This refers to a set of requirements
+  that MUST be satisfied in order to obtain
+  the plaintext data key.
```
robin-aws, line 132: This definition makes me think the name really should be some variation of "key decryption contract" to disambiguate.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
seebees: Why not "decryption contract" for the artifact and then it can be qualified by who it handling or verifying the contract?

changes/2020-06-04_how-to-fail-with-keyrings/background.md
```diff
@@ -0,0 +1,166 @@
...
+  - The encryption context MUST exactly match
+    the encryption context used on encrypt.
+
+- Multi-keyring:
+
+  - Satisfying any member keyring's decryption contract
+    MUST be sufficient to satisfy
+    the multi-keyring's decryption contract.
+
+## Issue 1 : When MUST a keyring fail?
```
SalusaSecondus, line 112: Is it reasonable to consider keyrings which are only one sided? Maybe many-to-one?

Many different key rings only "encrypt" but the resulting artifacts can all be decrypted by a single standardized keyring?

changes/2020-06-04_how-to-fail-with-keyrings/background.md
```diff
@@ -0,0 +1,166 @@
...
+this change affects the following keyring specifications.
+
+- Keyring interface :
+  The keyring interface defines no-op behavior as
+  "output[ing] the encryption/decryption materials unmodified".
+  This needs to be modified to frame this behavior as a failure
+  and state that encryption/decryption materials MUST NOT be modified
+  when a keyring fails.
+
+  - _This also affects all keyring specifications._
```
SalusaSecondus, line 144: I think there are two different failures cases:
1. I didn't see anything which matches my decryption contract and so I didn't even attempt anything. (NOP)
2. I saw something I thought I _could_ decrypt, but failed for some reason.

changes/2020-06-04_how-to-fail-with-keyrings/background.md
mattsb42-aws: 👍

changes/2020-06-04_how-to-fail-with-keyrings/change.md
robin-aws: As in the decryption contract for the artifact is implicitly a "key decryption contract" because of the kind of artifact it is, and so the decryption contract for the whole message is implicitly a "CMM decryption contract" or similar?

changes/2020-06-04_how-to-fail-with-keyrings/background.md
```diff
@@ -0,0 +1,166 @@
...
+
+  - On encrypt, if _any_ child keyring fails,
+    the multi-keyring MUST fail.
+  - On decrypt, if _all_ child keyrings fail,
+    the multi-keyring MUST fail.
+
+# One-Way Doors
+
+Once we define the decryption contract for a keyring,
+that contract MUST NOT change.
```
SalusaSecondus, line 160: It cannot become more restrictive. It can become more lenient.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
acioc: I personally prefer the `keyring decryption contract`. This is less about how to decrypt the key and more about what the keyring needs to be able to decrypt the key. For example, for the AWS KMS Keyring, you need AWS KMS Decrypt API permissions. The key itself doesn't care about network calls/ AWS permissions, but the keyring does care.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -113,111 +113,147 @@ in different environments and implementations.
...
-This set of requirements,
-manifested as the artifact,
-is what we call the "keyring decryption contract."
+a keyring creates one or more "decryption artifacts",
+sometimes called "encrypted data keys",
+that are included in the resulting encrypted message.
+On decryption,
+a keyring consumes a decryption artifact
+and fulfills a set of requirements
+by using the decryption artifact
```
seebees, line 122:
```suggestion
a keyring consumes decryption artifacts
and fulfills a set of requirements
by using these decryption artifacts
```

at a minimum the keyring gets multiple...

As I keep reading the plural singular is everywhere.
Maybe leaving it all singular is better.
But I don't want to require a 1:1 relationship.

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -113,111 +113,147 @@ in different environments and implementations.
...
 On encryption,
+a keyring creates one or more decryption artifacts.
+
+On decrypt,
+a keyring fulfills its decryption contract
+using one or more decryption artifacts.
+
+### How to Fail
+
+In encryption,
```
acioc, line 160:
```suggestion
On encryption,
```

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,259 @@
...
+### Definitions and Terms
+
+- **decryption artifact** :
+  A value that is consumed on decryption
+  to obtain decryption materials.
+
+- **decryption contract** :
+  A set of requirements that MUST be met
+  to obtain decryption materials
+  using a decryption artifact.
```
seebees, line 137:
```suggestion
  using one or more decryption artifacts.
```

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,259 @@
...
+
+- **creating the decryption artifact** :
+  The process of creating a decryption artifact
```
seebees, line 140:
```suggestion
  The process of creating one or more decryption artifacts
```

changes/2020-06-04_how-to-fail-with-keyrings/change.md
```diff
@@ -0,0 +1,259 @@
...
+  intended for a component with a specific decryption contract.
+
+- **fulfilling the decryption contract** :
+  The process of taking a decryption artifact
```
seebees, line 144:
```suggestion
  The process of taking one or more decryption artifacts
```

Issue #21, opened 2019-08-21
EDK Key Provider ID/Information field names


lavaleri:
Currently we use the terms "key provider ID" and "Key Provider Information" for fields of the EDK, however these terms don't exactly describe what we want. The implemented keyrings specifically produce EDKs that use these fields to store the "Key Name" and "Key Namespace".

Should we continue with the old "key provider ID" and "key provider information" terms, or should we update them to more accurately represent the data those fields should be capturing?

Acceptance criteria:
- Get team alignment on above question.
- If latter, get team alignment on name change and update spec to reflect new name.

mattsb42-aws:
also affects:

* #47

PR #135, opened 2020-06-12
docs: add change doc to describe key namespaces, names, provider IDs, and provider info

mattsb42-aws:
_Issue #, if available:_
resolves: #21 
resolves: #47 

_Description of changes:_

This change introduces definitions of the terms "key namespace", "key name", "provider ID", and "provider info", as well as  how they relate to each other.

**NOTE: I have not yet altered the keyring spec doc. That needs to happen before this merges.**

### DIFF
--- a/changes/2020-06-09_wrapping-key-identifiers/change.md
+++ b/changes/2020-06-09_wrapping-key-identifiers/change.md
@@ -88,9 +88,7 @@ This change SHOULD NOT have any security implications.
 
 ## Operational Implications
 
-If the keyring chooses to write a provider info value that does not equal its key name,
-that keyring's behavior can be more complex for customers to configure
-than if they are the same.
+This change SHOULD NOT have any operational implications.
 
 ## Guide-level Explanation
 
--- a/changes/2020-06-09_wrapping-key-identifiers/change.md
+++ b/changes/2020-06-09_wrapping-key-identifiers/change.md
@@ -92,7 +92,7 @@ This change SHOULD NOT have any operational implications.
 
 ## Guide-level Explanation
 
-Key namespace and key name are configuration values that MAY determine the behavior of a keyring.
+Key namespace and key name are configuration values that determine the behavior of a keyring.
 
 Provider ID and provider info are values that identify the keyring configuration
 that can fulfill the keyring's decryption contract.
@@ -100,7 +100,7 @@ The keyring attaches these values to the encrypted data key ciphertext.
 
 ## Reference-level Explanation
 
-"Key namespace", "key name", "provider ID", "provider info"
+"Key namespace", "key name", "provider ID", and "provider info"
 are all concepts that identify a wrapping key.
 
 ### key namespace
@@ -109,11 +109,15 @@ A configuration value for a keyring
 that identifies the grouping or categorization
 for the wrapping keys that keyring can access.
 
+The key namespace MUST be a string value.
+
 ### key name
 
 A configuration value for a keyring
 that identifies a single wrapping key
 within a key namespace.
+
+The key name MUST be a string value.
 
 ### provider ID
 
@@ -122,7 +126,8 @@ as part of an encrypted data key structure
 that identifies the grouping or categorization
 for the keyring that can fulfill this keyring's decryption contract.
 
-The provider ID SHOULD be equal to the key namespace.
+The provider ID MUST be a binary value
+and SHOULD be equal to a UTF-8 encoding of the key namespace.
 
 ### provider info
 
@@ -131,14 +136,15 @@ as part of an encrypted data key structure
 that provides necessary information for a keyring
 to fulfill this keyring's decryption contract.
 
-The provider info SHOULD be equal to the key name.
+The provider info MUST be a binary value
+and SHOULD be equal to a UTF-8 encoding of the key name.
 
 One example of a keyring where the key name and provider info can differ
 is the AWS KMS keyring.
 This keyring uses its key name to identify the desired CMK in its call to AWS KMS.
 However, the provider info that this keyring writes is
 the CMK identifier that AWS KMS includes in its response.
-If the key name is a CMK ARN these two values are identical
+If the key name is a CMK ARN, these two values are identical
 because this response value is always the CMK ARN of the CMK that AWS KMS used.
 However, if the key name is some other valid CMK identifier,
 such as an alias,
--- a/changes/2020-06-09_wrapping-key-identifiers/change.md
+++ b/changes/2020-06-09_wrapping-key-identifiers/change.md
@@ -124,7 +124,7 @@ The key name MUST be a string value.
 An output value returned by a keyring on encrypt
 as part of an encrypted data key structure
 that identifies the grouping or categorization
-for the keyring that can fulfill this keyring's decryption contract.
+for a keyring that can fulfill this decryption contract.
 
 The provider ID MUST be a binary value
 and SHOULD be equal to a UTF-8 encoding of the key namespace.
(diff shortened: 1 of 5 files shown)


changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,147 @@
...
+terms that already exist
+and the existing relationship between them.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+If the keyring chooses to write a provider info value that does not equal its key name,
```
lavaleri, line 91: nit: this doesn't appear to be an operational implication for defining the terms "key name" and "provider info", this is just operational implications of how these concepts as they currently exist (which we aren't changing).

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,145 @@
...
+The provider ID SHOULD be equal to the key namespace.
+
+### provider info
+
+An output value returned by a keyring on encrypt
+as part of an encrypted data key structure
+that provides necessary information for a keyring
+to fulfill this keyring's decryption contract.
+
+The provider info SHOULD be equal to the key name.
```
MatthewBennington, line 134: Note: "provider info", and "key name" are defined with slightly different types.

"provider info" is defined as "usually a UTF-8 encoded value", where as "key name" is defined specifically as a "UTF-8 encoded value" in both raw keyrings.

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,145 @@
...
+
+One example of a keyring where the key name and provider info can differ
```
seebees, line 136: It the Raw AES keyring another good example?

changes/2020-06-09_wrapping-key-identifiers/change.md
seebees: This is true, and ideally we would have some _why_ written down about why "provider info" is bytes and not a string

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,145 @@
...
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change SHOULD NOT have any operational implications.
+
+## Guide-level Explanation
+
+Key namespace and key name are configuration values that MAY determine the behavior of a keyring.
```
seebees, line 95: There is something about this MAY that makes me a little uncomfortable.
Like a keyring that does not delegate, may not even have these values.
Below you talk about how the values SHOULD be equal.
Perhaps here is a place to express the relationship more concretely?

I understand that
the "key namespace" and the "key name" as "for the keyring"
and the "provider ID and "provider info" as "for the encrypted data key".

changes/2020-06-09_wrapping-key-identifiers/change.md
mattsb42-aws: It's an example...I don't know that it's a *good* example. :/

I'd rather not call attention to it shoving extra stuff into the provider info because that's a pattern that we do not want to encourage in the future.

seebees: totally fair

mattsb42-aws:
Thinking this over again today, I've gone with "key name/namespace are always strings" and "provider ID/info are always bytes", and adjusted accordingly.

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,151 @@
...
+within a key namespace.
+
+The key name MUST be a string value.
+
+### provider ID
+
+An output value returned by a keyring on encrypt
+as part of an encrypted data key structure
+that identifies the grouping or categorization
+for the keyring that can fulfill this keyring's decryption contract.
```
seebees, line 127:
```suggestion
for a keyring to fulfill this its decryption contract.
```

From #131, I'm still wondering about where
the decryption contract is.
I _think_ that the contract **is** the encrypted data key?
But then what do we say about keyrings,
that that can unwrap the encrypted data key
but chose not to?

mattsb42-aws: > But then what do we say about keyrings,
that that can unwrap the encrypted data key
but chose not to?

In that scenario I would say that is an example of a keyring being "unwilling" (as opposed to "unable") to fulfill the decryption contract.

framework/raw-aes-keyring.md
```diff
@@ -50,22 +50,10 @@ Galois/Counter Mode (GCM) Specification: [NIST Special Publication 800-38D](http
...
-A UTF-8 encoded value that, together with the [key name](#key-name), identifies a particular [wrapping key](#wrapping-key).
-
-This value is also used for bookkeeping purposes.
-
-### Key Name
-
-A UTF-8 encoded value that, together with the [key namespace](#key-namespace), identifies a particular [wrapping key](#wrapping-key).
-
-This value is also used for bookkeeping purposes.
-
```
MatthewBennington, line 68: The equivalent lines should be removed from raw-rsa-keyring.

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,151 @@
...
+
+### key name
+
+A configuration value for a keyring
+that identifies a single wrapping key
+within a key namespace.
+
+The key name MUST be a string value.
+
+### provider ID
```
robin-aws, line 122: Is this implicitly a "key provider ID" and should we qualify it as such, or are you intentionally trying to disambiguate it from the "key provider ID" in an EDK?

framework/raw-aes-keyring.md
mattsb42-aws: I'm pretty sure those sections were not present in raw-rsa-keyring when I wrote this. It looks like they are in master now; addressed in merge from master.

changes/2020-06-09_wrapping-key-identifiers/change.md
mattsb42-aws: I forgot we had that named with "key". I'll update and link to the structures value.

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,151 @@
...
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                                                                             |
+| --------------------------------------------------------------------------------------------------- |
+| [Encrypted Data Key](../../framework/structures.md#encrypted-data-key)                              |
+| [AWS KMS Keyring](../../framework/kms-keyring.md)                                                   |
+| [Raw AES Keyring](../../framework/raw-aes-keyring.md)                                               |
+| [Raw RSA Keyring](../../framework/raw-rsa-keyring.md)                                               |
+| [Keyring Decryption Contract](https://github.com/awslabs/aws-encryption-sdk-specification/pull/131) |
```
MatthewBennington, line 16: @acioc had the idea of making these permanent links to the specific commit before the change.

changes/2020-06-09_wrapping-key-identifiers/change.md
```diff
@@ -0,0 +1,151 @@
...
+
+This change SHOULD NOT have any operational implications.
+
+## Guide-level Explanation
+
+Key namespace and key name are configuration values that determine the behavior of a keyring.
+
+Provider ID and key provider info are values that identify the keyring configuration
+that can fulfill the keyring's decryption contract.
+The keyring attaches these values to the encrypted data key ciphertext.
```
MatthewBennington, line 99:
```suggestion
The keyring attaches these values to a data key ciphertext to form an encrypted data key.
```

changes/2020-06-09_wrapping-key-identifiers/change.md
acioc: I think there's some ongoing discussion of this on https://github.com/awslabs/aws-encryption-sdk-specification/issues/137

Issue #102, opened 2020-05-06
Clarify Encrypt behavior based on plaintext length input


lavaleri:
Related to unblocking: #71 

### Encryption "plaintext length" param behavior

When performing the encrypt operation on a plaintext where the plaintext length is not immediately known, users MUST be able to specify the optional parameter `maxPlaintextLength`. The value of this field represents the bound on the length of the plaintext to be encrypted. The ESDK MUST NOT encrypt a plaintext greater than this length, and MUST fail if it can be determined during encryption that the actual plaintext length is greater than what the user supplied on input. The actual name of this parameter MAY be different per implementation.

Similarly, the CMM's GetEncryptionMaterials should clearly define `maxPlaintextLength` as an input that represents the maximum size of the plaintext that will be encrypted using these materials. Having this parameter is integral for having the Caching CMM work for streaming encrypt operations. The actual name of this input MAY be different per implementation.

However, what should happen if a user inputs a plaintext with a known length (e.g. a byte array instead of a stream) as well as inputs a (possible incorrect) `maxPlaintextLength`?

We should avoid letting users do this as much as is reasonable for the implementation. If for a certain API the supplied plaintext's length will always be known, that API shouldn't provide `maxPlaintextLength` as an input. As such, we should add the following to the spec:
- If exposing an API where the plaintext length is always known from the input plaintext, then the ESDK MUST NOT provide `maxPlaintextLength` as a user input.
- If exposing an API where it is possible for the length of the plaintext to be unknown on input, then the ESDK MUST provide an optional `maxPlaintextLength` input.

To make clear the behavior in the latter case, we should add the following to the spec:
- If `maxPlaintextLength` is supplied on input and the input plaintext has a total length greater than `maxPlaintextLength`, this operation MUST fail.
- If the length of the plaintext is known, the Encrypt operation MUST pass the real plaintext length as `maxPlaintextLength` in the CMM GetEncryptionMaterials call. (regardless of whether maxPlaintextLength was supplied on input),
- If the length of the plaintext is not known and `maxPlaintextLength` was supplied on input, the Encrypt operation MUST pass the supplied `maxPlaintextLength` to the CMM GetEncryptionMaterials call.
- If the length of the plaintext is not known and `maxPlaintextLength` was not supplied on input, the Encrypt operation MUST NOT specify a `maxPlaintextLength` in the CMM GetEncryptionMaterials call.

[...]

PR #138, opened 2020-06-17
doc: add change doc to specify max plaintext length input on encrypt

lavaleri:
_Issue #, if available:_ Related to #102 and #71

_Description of changes:_ Add change document to describe spec update for specifying the input on Encrypt that puts a bound on the length of the input plaintext.


changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+| Specification                                     |
+| ------------------------------------------------- |
+| [Encrypt](../../client-apis/encrypt.md)           |
+| [CMM Interface](../../framework/cmm-interface.md) |
+
+## Affected Implementations
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| Python     | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-python)              |
```
alex-chew, line 24: I think this label should read "aws-encryption-sdk-python".

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+## Summary
+
+When performing the encrypt operation on an unknown length plaintext,
+users MUST be able to specify some input that bounds the length of the plaintext to be encrypted.
+
+This input already exists in the specification as [plaintext length](../../client-apis/encrypt.md#plaintext-length)
+however it's effect on the Encrypt operation's behavior is unspecified.
+
+This change renames this input in the specification
+to be a more accurate reflection of it's behavior,
```
alex-chew, line 60: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+For example, if a user supplies plaintext as a stream to Encrypt,
+that is a unknown-length plaintext.
+
+## Summary
+
+When performing the encrypt operation on an unknown length plaintext,
+users MUST be able to specify some input that bounds the length of the plaintext to be encrypted.
+
+This input already exists in the specification as [plaintext length](../../client-apis/encrypt.md#plaintext-length)
+however it's effect on the Encrypt operation's behavior is unspecified.
```
alex-chew, line 57: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+
+- How different implementations may allow a user to specify this input for the Encrypt operation
+  is out of scope.
+- Significantly changing the shape of any of our Encrypt API implementations is out of scope.
+
+## Motivation
+
+All implementations allow users to specify a plaintext length when calling Encrypt with a
+unknown-length plaintext.
+However, it's behavior in the Encrypt operation is not clearly specified.
```
alex-chew, line 73: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+
+Additionally, some implementations also allow a plaintext length to be specified on Encrypt with a
+known-length plaintext.
+The behavior of Encrypt in this case is not consistent between implementations.
+
+The purpose of this change is to document the correct behavior of this input and
+rename this input to clarify it's intent.
```
alex-chew, line 80: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+This change will break any JS user that is currently using Encrypt with a known-length plaintext
+and inputting a plaintextLength less than that length.
+
+This change will break any Python user that depends on the source_length being sent to the
+CMM instead of the true length of a known-length plaintext.
+
+## Guide-level Explanation
+
+This change renames the `plaintext length` Encryption input to `max plaintext length` in the specification,
+and clarifies it's behavior in the Encrypt operation.
```
alex-chew, line 123: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+
+This change also renames the `plaintext length` field in the CMM interface to `max plaintext length`.
+
+Finally, this change specifies when this input should be available for users to specify:
+if and only if the input plaintext can be of unknown-length.
+
+## Reference-level Explanation
+
+We need to update the spec to make the purpose and behavior of the `plaintext length` input clear
+by renaming it and specifying it's exact behavior within the Encrypt operation:
```
alex-chew, line 143: nit: its

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,179 @@
...
+
+Any plaintext input to the Encrypt operation where the total length of that plaintext
+cannot be immediately determined.
+For example, if a user supplies plaintext as a stream to Encrypt,
+that is a unknown-length plaintext.
+
+## Summary
+
+When performing the encrypt operation on an unknown length plaintext,
+users MUST be able to specify some input that bounds the length of the plaintext to be encrypted.
```
alex-chew, line 54: Should the summary also state that users MUST NOT be able to specify the maximum plaintext length when performing Encrypt on a known-length plaintext?

lavaleri: We need to be careful about the wording, but yes. I've updated the summary.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,182 @@
...
+without significantly changing this input to benefit the customer.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change will break any JS user that is currently using Encrypt with a known-length plaintext
+and inputting a plaintextLength less than that length.
```
seebees, line 118: No, the 2 JS implementations do not let this happen.
The browser does not accept the option at all because **all** operations are one-shot.
https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/encrypt-browser/src/encrypt.ts#L40
The Node implementation will clobber any value provided in favor of the correct value. 
https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/encrypt-node/src/encrypt.ts#L30

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,182 @@
...
+Additionally, when calling the CMM's GetEncryptionMaterials call during the Encrypt operation,
+the operation MUST pass in the length of the input plaintext if it can be known.
+If it can't be known, the operation passes in `max plaintext input` if supplied.
+If it can't be known and no `max plaintext input` is supplied, the operation supplies no value for
+this field.
+
+This change also renames the `plaintext length` field in the CMM interface to `max plaintext length`.
+
+Finally, this change specifies when this input should be available for users to specify:
+if and only if the input plaintext can be of unknown-length.
```
seebees, line 141: How should this be accomplished?
The JS encrypt function takes multiple options.
JavaScript does not have any function overloads.

Should there be multiple `encrypt` functions?
What should they be called?
Or is what it is doing currently adequate?

What does "available" mean?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,182 @@
...
+so we should describe it in the specification as such.
+
+This change specifies that the Encrypt operation MUST pass the real length of the plaintext
+to the CMM if known instead of a user supplied bound because the former is more accurate.
+There is no case where a user would benefit from passing
+a less accurate `max plaintext length` to the CMM.
+If there were such a case,
+then such a CMM is not correctly using the intent behind `maxPlaintextLength`.
+It is better to restrict flexibilty here so that users have less opportunity
+to depend on behaviors from CMMs that are misusing this value.
```
seebees, line 101: To be clear: Using this behavior to try and flush the Caching CMM is not a good idea.
Say, trying to encrypt 0 bytes with a plaintext length of 2**66 to ensure
that the next encrypt gets a new data key.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,182 @@
...
+When performing the encrypt operation on a unknown-length plaintext,
+users MUST be able to specify an optional parameter `maxPlaintextLength`.
+The value of this field represents the bound on the length of the plaintext to be encrypted.
+The ESDK MUST NOT encrypt a plaintext greater than this length,
+and MUST fail if it can be determined during encryption that the actual plaintext length
+is greater than what the user supplied on input.
+The actual name of this input, and how the user specifies this value for the Encrypt operation
+MAY be different per implementation.
+
+We also need to specify the exact behavior for how this input is used in the CMM's GetEncryptionMaterials call:
```
seebees, line 157:
```suggestion
We also need to specify the exact behavior for how this input is passed to the CMM's GetEncryptionMaterials call:
```

I don't think you mean "used" because you don't talk about how the CMM is using the value.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
lavaleri: I will update to specify "Node JS". The Node JS implementation violates the spec I've outlined here. Instead of clobbering in the "less than" case, it MUST fail. Therefore, anyone who was previously depending on the implementation just using the correct value in the "less than" case will break.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
lavaleri: That is a more concrete example of what I'm trying to avoid. Users should not be using this value to be triggering weird behaviors in the CMM. Maybe framing as "the CMM is doing a bad thing" isn't the correct way to frame it though, and instead say "users shouldn't expect setting this value to change the CMM's behavior".

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
lavaleri: I think answers to these questions are more clearly addressed in the "Reference-level Explanation" section. I was actually a bit unsure what detail to go into in this one. As it is right now, this sentence is not intending to specify what the exact shape of the API should look like, only that if you provide an API that "only" takes in known-length plaintexts, then it doesn't make sense for users to be able to specify this input.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
seebees: Hum, if what you want is to "not let it be an option" then how would that work, if not this way?

lavaleri:
After discussion with @seebees offline, we determined that it either makes sense to interpret `max plaintext size` on both Encrypt and GetEncryptionMaterials as the same control, and the control on Encrypt representing a passthrough that must also be available to known-length plaintexts (and thus we should pass the user intent to the CMM)

OR

This is a control that MUST only be considered for unknown-length plaintexts, in which case, how an implementation handles funneling user intent into that valid set of inputs for Encrypt depends on the language. We shouldn't thus say that all implementations MUST fail if given a known-length plaintext with a `plaintext length bound` less than the real length. Instead, we describe the only use case in which using `plaintext length bound` is valid within the context of Encrypt.

I have gone for the latter and updated the language in this doc to reflect that change, and scope the discussion to the Encrypt operation.

Additionally, for this discussion the Guide v. Reference level discussion was confusing given that it is mostly talking about changing the spec. Since i didn't find discussing the changes on these two levels useful, I'm considering the Reference level discussion casual enough to be the Guide level discussion. If you think there are better ways to go about this, let me know.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Guide-level Explanation
+
+We need to update the spec to make the purpose and behavior of the `plaintext length` input clear
+by renaming it and specifying its exact behavior within the Encrypt operation:
+
+When performing the Encrypt operation on an unknown-length plaintext,
+users MUST be able to specify an optional parameter `plaintext length bound`.
+The value of this field represents the max length of the plaintext to be encrypted.
+The ESDK MUST NOT encrypt a plaintext greater than this length,
```
seebees, line 164: I feel like this is a little buried.  I can see why it is not in the list below, but I think that a little higher billing is in order.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+that is a unknown-length plaintext.
+
+## Summary
+
+The Encrypt operation takes a plaintext as input that can either have a known length or an
+unknown length.
+If the input plaintext is a unknown-length plaintext,
+the Encrypt operation MUST also take an optional input that bounds the length of that input.
+
+The specification already specifies a [plaintext length](../../client-apis/encrypt.md#plaintext-length)
```
acioc, line 57: Should this be an absolute link (with the SHA)? So a different header in the future still works for the link?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+input on the Encrypt operation that describes a bound on the plaintext length,
+however its effect on the Encrypt operation's behavior is unspecified.
+
+This change renames this input to `plaintext length bound` in the specification
+to be a more accurate reflection of its behavior and intent,
+and specifies the correct behavior during the Encrypt operation.
+Specifically, in the scope of the Encrypt operation it's value MUST be passed to the CMM.
```
acioc, line 64: np: `its`

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+Otherwise, the Encrypt operation MUST pass this value to the CMM.
+
+The intention of this value is not to be a one to one passthrough value to the CMM.
+If that were the case,
+then we would be concerned with always letting users specify this value for known-length plaintexts,
+or with possibly letting users pass in a "unknown length" intent.
+We do not want to support these cases.
+Users should not be setting a `max plaintext length` value for known-length plaintexts
+with the intention of changing CMM behavior.
+Instead, users should be specifying this value in for unknown-length plaintexts in order to use CMMs
```
acioc, line 113: Users `should`/`should not` or users `must`/`must not`?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+This is the current state of the behavior, so the specification should describe it as such.
+This is renamed to a different name than `plaintext length bound` in order to
+distinguish the two controls.
+
+This change does not strictly prescribe how users should express intent for the plaintext
+and a possible `plaintext length bound` through an API.
+How an API should be designed with these controls in mind depends greatly on the capabilities and
+idioms of the language in use.
+Specifically, this gets tricky for languages where the idiom is to
+accept an object represent a set of user supplied params.
```
acioc, line 129: accept an object that represents a set?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+If a user supplies both a known-length plaintext and a `max plaintext length`,
+what should be done?
+It doesn't make sense to have to check for bad values in the user supplied params,
+given it's unstructured nature.
+Thus, our specification reccomends that languages SHOULD either ensure that `max plaintext length`
```
acioc, line 134: np: `recommends`

np: `languages SHOULD either ensure ` -> `languages SHOULD ensure ` ?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,203 @@
...
+isn't supplied with a known-length plaintext by construction.
+If it doesn't, then it SHOULD ignore any user supplied `max plaintext length` in the case that
+a known-length plaintext is supplied.
+
+## Drawbacks
+
+This change does not prescribe the name of this input in implementations,
+and thus there will not be a standardized name or way of specifying this input across implementations.
+This is because this control already exists in implementations under different names,
+and it is not worth breaking all customers to rename this input
```
acioc, line 144: np: `breaking all customers` -> `breaking customer-facing APIs`

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
lavaleri: The first is definitely a SHOULD NOT, which makes the latter be a SHOULD. Alternatively, I'd be ok with using a different verb to avoid RFC 2119 verbs when not strictly necessary.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
lavaleri: Where do you think this should be highlighted more? In the summary? Reference it again later?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,204 @@
...
+
+We need to update the spec to make the purpose and behavior of the `plaintext length` input clear
+by renaming it and specifying its exact behavior within the Encrypt operation:
+
+When performing the Encrypt operation on an unknown-length plaintext,
+customers MUST be able to specify an optional parameter `plaintext length bound`.
+The value of this field represents the max length of the plaintext to be encrypted.
+The ESDK MUST NOT encrypt a plaintext greater than this length,
+and MUST fail if it can be determined during encryption that the actual plaintext length
+is greater than what the customer supplied on input.
```
seebees, line 167:
```suggestion
The ESDK MUST NOT encrypt a plaintext greater than this length.
If it is determined during encryption that the actual plaintext length
is greater than what the customer supplied on input
the ESDK MUST fail.
```

client-apis/encrypt.md
```diff
@@ -75,10 +84,13 @@ The [algorithm suite](#algorithm-suite.md) that SHOULD be used for encryption.
 The [frame length](../data-format/message-header.md#frame-length) to use for [framed data](../data-format/message-body.md).
 This value MUST NOT exceed the value 2^32 - 1.
 
-### Plaintext Length
+### Plaintext Length Bound
 
 A bound on the length of the [plaintext](#plaintext) to encrypt.
```
seebees, line 89:
```suggestion
A bound on the length of [plaintext](#plaintext) with an unknown length to encrypt.
```
I think we should be clear here that this value is for.

client-apis/encrypt.md
```diff
@@ -99,6 +111,11 @@ MUST be constructed as follows:
   Otherwise, this is an empty encryption context.
 - Algorithm Suite: If provided, this is the [input algorithm suite](#algorithm-suite).
   Otherwise, this field is not included.
+- Max Plaintext Length: If the [input plaintext](#plaintext) has known length, this is the
+  length of the plaintext.
```
seebees, line 115:
```suggestion
- Max Plaintext Length: If the [input plaintext](#plaintext) has known length,
  this length MUST be used.
```

client-apis/encrypt.md
```diff
@@ -153,6 +170,10 @@ Each frame of the [message body](../data-format/message-body.md) is serialized w
   - The plaintext contains part of the input [plaintext](#plaintext) this frame is encrypting.
 - [Authentication Tag](../data-format/message-body.md#authentication-tag): MUST be the authentication tag outputted by the above encryption.
 
+If [Plaintext Length Bound](#plaintext-length-bound) was specified on input
+and this operation can determine that the plaintext being encrypted has a length greater than this value,
```
seebees, line 174:
```suggestion
and this operation determines that the plaintext being encrypted has a length greater than this value,
```
How could the operation not determine this?
This also resolves my "top billing comment" :)

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,205 @@
...
+
+Additionally, it is unclear in the spec what the behavior should be
+if the input plaintext has a known-length
+and the customer specifies a bound on the length of that plaintext.
+
+This change proposes that in the scope of the Encrypt operation,
+an input plaintext is either of known length,
+of unknown length with customer supplied estimated size,
+or unknown length with no customer supplied estimated size.
+The Encrypt operation should never consider a `plaintext length bound` alongside
```
mattsb42-aws, line 75: should not consider or should not allow?

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,205 @@
...
+or with possibly letting customers pass in a "unknown length" intent.
+We do not want to support these cases.
+Customers SHOULD NOT set a `max plaintext length` value for known-length plaintexts
+with the intention of changing CMM behavior.
+Instead, customers SHOULD specify this value for unknown-length plaintexts in order to use CMMs
+that depend on plaintext length for their internal logic.
+
+This change additionally renames the `plaintext length` field on GetEncryptionMaterialsRequest to
+`max plaintext length` because that better explains the intent behind this field.
+This field describes the max size of the plaintext which will be encrypted using the materials
```
mattsb42-aws, line 119:
```suggestion
This field describes the max size of the plaintext that will be encrypted using the materials
```

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,205 @@
...
+then we would be concerned with always letting customers specify this value for known-length plaintexts,
+or with possibly letting customers pass in a "unknown length" intent.
+We do not want to support these cases.
+Customers SHOULD NOT set a `max plaintext length` value for known-length plaintexts
+with the intention of changing CMM behavior.
+Instead, customers SHOULD specify this value for unknown-length plaintexts in order to use CMMs
+that depend on plaintext length for their internal logic.
+
+This change additionally renames the `plaintext length` field on GetEncryptionMaterialsRequest to
+`max plaintext length` because that better explains the intent behind this field.
```
mattsb42-aws, line 118: I've read this through several times now and I just get more confused each time. :/

What is the difference between `plaintext length bound` and `max plaintext length` other than one being part of the Encrypt API and one being part of the CMM GetEncryptionMaterials API?

lavaleri: `max plaintext length` is an input on the CMM GetEncryptionMaterials API. It describes the maximum size of the plaintext that the CMM caller intends to encrypt using the result materials. I assert that in the context of Encrypt, if the plaintext has a known length this MUST be exactly the length of the plaintext.

`plaintext length bound` is an input on Encrypt. If provided, it describes the maximum size of a unknown-length plaintext. It specifically MUST NOT describe the intent of "I, as a caller, want the Encrypt operation to pass this value in to the CMM instead of the actual length."

They both describe "plaintext MUST NOT be larger than this size", but the intentions are subtly different. In order to distinguish them, I have given them slightly different names.

changes/2020-06-17_encrypt-max-plaintext-length-input/change.md
```diff
@@ -0,0 +1,205 @@
...
+Specifically, this gets tricky for languages where the idiom is to
+accept an object that represents a set of customer supplied params.
+If a customer supplies both a known-length plaintext and a `max plaintext length`,
+what should be done?
+It doesn't make sense to have to check for bad values in the customer supplied params,
+given its unstructured nature.
+Thus, our specification recommends that languages SHOULD ensure that `max plaintext length`
+isn't supplied with a known-length plaintext by construction.
+If it doesn't, then it SHOULD ignore any customer supplied `max plaintext length` in the case that
+a known-length plaintext is supplied.
```
alex-chew, line 138: Assuming the guide-level explanation below is correct, then this should be a MUST rather than a SHOULD.

mattsb42-aws:
I like the intention of this change.
I still find it confusing that we have two similar but different controls with two similar but different names, but I won't block on that.

PR #140, opened 2020-06-18
doc(Raw Keyrings): Raw keyrings MUST NOT accept a key namespace of "aws-kms"

### LINKED ISSUE #101: Raw keyrings MUST NOT accept "aws-kms" key namespace
To protect the namespace of the AWS KMS keyring, the raw AES keyring and the raw RSA keyring MUST not allow a key namespace of `aws-kms`.

MatthewBennington:
_Issue #, if available:_ Resolves #101 

_Description of changes:_
- Clarify that Raw keyrings (raw AES keyring and raw RSA keyring) "MUST NOT" accept a key namespace value of `"aws-kms"`
- Bump the minor version of raw AES keyring and raw RSA keyring

### DIFF
--- a/framework/raw-aes-keyring.md
+++ b/framework/raw-aes-keyring.md
@@ -5,9 +5,13 @@
 
 ## Version
 
-0.2.0
+0.3.0
 
 ### Changelog
+
+- 0.3.0
+
+  - [Raw AES keyring MUST NOT accept a key namespace of "aws-kms".](https://github.com/awslabs/aws-encryption-sdk-specification/issues/101)
 
 - 0.2.0
 
@@ -57,6 +61,8 @@ On keyring initialization, the following inputs are REQUIRED:
 ### Key Namespace
 
 A UTF-8 encoded value that, together with the [key name](#key-name), identifies a particular [wrapping key](#wrapping-key).
+
+The raw AES keyring MUST NOT accept a key namespace of "aws-kms".
 
 This value is also used for bookkeeping purposes.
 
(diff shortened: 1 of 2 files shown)


framework/raw-aes-keyring.md
```diff
@@ -5,10 +5,14 @@
...
 ## Version
 
-0.2.0
+0.3.0
 
 ### Changelog
 
+- 0.3.0
+
+  - [Raw AES keyring MUST NOT accept a key namespace of "aws-kms".](https://github.com/awslabs/aws-encryption-sdk-specification/issues/101)
```
acioc, line 14: np: probably don't need a `.` here

framework/raw-rsa-keyring.md
```diff
@@ -57,10 +61,16 @@ On keyring initialization, the following inputs are REQUIRED:
 
 A UTF-8 encoded value that namespaces this keyring.
 
+The raw RSA keyring MUST NOT accept a key namespace of "aws-kms".
+
+This value is also used for bookkeeping purposes.
```
acioc, line 66: This is probably fine (since it looks like it was also part of the Raw AES keyring). But always good to clarify this further. Does "bookkeeping" mean "auditing" or something else in this context?

seebees: I think we should look at #135 for how we use these.
I think that bookkeeping is too weak :)
I think that they exists _so that_ the keyring can recognize the encrypted data keys on decrypt.
That is not just bookkeeping (I think?)

seebees: Hum thinking more, I'm inclined to keep this so the current spec is consistent.
But I think that #135 or a followup should come back and address this.

Issue #103, opened 2020-05-07
add "required examples" specification


mattsb42-aws:
Add a specification doc that details the required examples, as laid out here[1].

Things we need to define include:

* The examples readme structure.
  * This is the structure and content laid out here[1].
* The list of all example scenarios.
  * This is the complete list of examples defined here[1].
* Requirements for any given example.
  * Every example MUST be independent.
  * Every example MUST NOT require additional or prior knowledge to understand.
  * Every example MUST explain step by step what it is doing.
  * Those explanations MUST differentiate between things that we are recommending that customers do vs things that we are demonstrating to customers.

[1] https://github.com/aws/aws-encryption-sdk-python/tree/master/examples

robin-aws:
Another less obvious example to include: demonstrate that it is possible and encouraged to create multiple keyrings/master key providers in a single application. Our existing examples demonstrate encrypting a single message with multiple CMKs, but I haven't found any that demonstrate creating a separate keyring for each customer according to the context. We could label this a "multi-tenancy" example.

PR #143, opened 2020-06-25
docs: add required examples

mattsb42-aws:
_Issue #, if available:_

#103 

_Description of changes:_

This initial commit only contains the requirements for examples. Either through expanding this PR or via a separate PR, I'll be adding spec documents for each of the required examples.

### DIFF
--- a/changes/2020-06-16_required-examples/change.md
+++ b/changes/2020-06-16_required-examples/change.md
@@ -18,8 +18,8 @@ This serves as a reference for all implementations that this change affects.
 
 | Language   | Repository                                                                            |
 | ---------- | ------------------------------------------------------------------------------------- |
-| Python     | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-python)              |
-| Java       | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-java)       |
+| Python     | [aws-encryption-sdk-python](https://github.com/aws/aws-encryption-sdk-python)              |
+| Java       | [aws-encryption-sdk-java](https://github.com/aws/aws-encryption-sdk-java)       |
 | C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
 | Javascript | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-javascript) |
 
@@ -45,11 +45,17 @@ The template document for each example contains those details.
 
 ## Motivation
 
-The two primary motivations for requiring examples are
-to provide guidance to AWS Encryption SDK customers
-and to force AWS Encryption SDK developers
-to use the APIs that they design and build
-to verify that they provide the intended user experience.
+The primary motivations for requiring examples include:
+
+- Providing guidance to AWS Encryption SDK customers.
+- Facilitating documentation and interoperability.
+- Encouraging AWS Encryption SDK developers
+  to work backwards from the caller experience
+  to verify that APIs and components
+  provide the intended functionality,
+  clarity, and ease of use.
+- Demonstrating that all implementations
+  can be used to solve the same problems.
 
 ## Drawbacks
 
@@ -90,8 +96,10 @@ the implementation's continuous integration testing.
 
 - Every example MUST only contain logic that is reasonable for production use.
 - If an example MUST contain logic that is not reasonable for production use,
-  it MUST include clear comments identifying that logic as such
-  and instructing the reader what they SHOULD do instead.
+  it MUST include clear comments that
+  MUST identify that logic as not fit for production use,
+  SHOULD explain why it is not fit for production use,
+  and MUST instruct the reader what they SHOULD do instead.
 
   - ex: Raw keyring examples generate wrapping keys as part of the example.
     These examples MUST contain guidance that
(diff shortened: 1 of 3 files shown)


changes/2020-06-16_required-examples/change.md
```diff
@@ -0,0 +1,98 @@
...
+the implementation's continuous integration testing.
+
+## Reference-level Explanation
+
+- Every example MUST be tested
+  as part of the implementation's continuous integration testing.
+- Every example MUST be independent and complete.
+- Every example SHOULD follow a consistent layout and framing.
+
+  - ex: Consistent function name and input parameters across examples.
```
acioc, line 89: This is fine since it's a SHOULD, but I definitely expect different implementations to have language-specific naming and (potentially) a different number of parameters for specific operations.

examples/examples.md
```diff
@@ -0,0 +1,52 @@
...
+### Changelog
+
+- 0.1.0
+
+  - [Require examples](../changes/2020-06-16_required-examples/change.md)
+
+## Implementations
+
+- [Python (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/)
+- [Java (DEV)](https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/examples/)
```
acioc, line 19: (DEV)? Both here and below.

changes/2020-06-16_required-examples/change.md
mattsb42-aws: Yup. My intention there was to avoid requiring the same pattern across implementations but to encourage each implementation to be internally consistent.

examples/examples.md
mattsb42-aws: That's meant to leave a breadcrumb for the existing implementations but explicitly remind us that they're not in master yet.

mattsb42-aws:
If there are no objections, I think that introducing additional templates for specific examples fits best as a separate PR.

changes/2020-06-16_required-examples/change.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Every example MUST only contain logic that is reasonable for production use.
+- If an example MUST contain logic that is not reasonable for production use,
+  it MUST include clear comments identifying that logic as such
+  and instructing the reader what they SHOULD do instead.
```
alex-chew, line 94: Should the comments also identify the reason that the logic is not reasonable for production use?

changes/2020-06-16_required-examples/change.md
```diff
@@ -0,0 +1,98 @@
...
+are out of scope for this document.
+The template document for each example contains those details.
+
+## Motivation
+
+The two primary motivations for requiring examples are
+to provide guidance to AWS Encryption SDK customers
+and to force AWS Encryption SDK developers
+to use the APIs that they design and build
+to verify that they provide the intended user experience.
```
seebees, line 52: This is the same thing,
but I also point out that
this helps us work backwards
from the customers wanting to use the tool.

If our process of development includes
generating the examples
we are more likely to build something
that customers will successfully use.

Maybe this is just here for posterity,
but it seems like an additional nice feature
of having examples.

changes/2020-06-16_required-examples/change.md
```diff
@@ -0,0 +1,98 @@
...
+## Out of Scope
+
+The specific implementations details for each example
+are out of scope for this document.
+The template document for each example contains those details.
+
+## Motivation
+
+The two primary motivations for requiring examples are
+to provide guidance to AWS Encryption SDK customers
```
seebees, line 49: Coming back to this,
I also like that this facilitates documentation
and the interoperability.

Customers can expect that the things
generally work the same way everywhere.
As with the comment below,
this is clearly true,
but may be worth explicitly stating

changes/2020-06-16_required-examples/change.md
```diff
@@ -0,0 +1,98 @@
...
+Requiring each implementation to include many examples
+increases the level of effort to create each implementation.
+We believe that the value to customers of
+improving the usability of APIs
+and providing guidance on how to use them
+justifies this cost.
+
+## Security Implications
+
+We MUST assume that any example will be used in production.
```
seebees, line 65: Since these example MUST run under CI,
we should be explicit that all keys
come from the test vectors.

mattsb42-aws: I'm hesitant to blend examples and test vectors, mainly because it feels like it is over-complicating matters. I could see the argument from the perspective of making it even more clear that you need to do *something* to manage your wrapping keys in production, but at the same time I worry that it might go the other direction and come across as "here is an example of *how* you might manage your wrapping keys", which is explicitly not what we are trying to show.

seebees: Ok, then what is at stake for me, is minimizing the number of wrapping keys that are floating around. Since we have to assume that this code will be copied and used in production, I want to minimize peoples ability to accidentally copy an RSA private key...

mattsb42-aws: This is one reason why for the current example implementations we generate a new wrapping key inside the example[1]. Hopefully we have made it sufficiently clear in the comments that you need to get your wrapping keys from somewhere else in production, but worst case if someone does copy-paste into production, the failure mode is data loss (because the wrapping key is never stored) rather than data compromise (because a publicly accessible wrapping key is used). Neither of these things is *good* but data loss is *less bad*.

[1] https://github.com/aws/aws-encryption-sdk-python/blob/5628ede2564f1b2003088d7f3901cda87ce0a0de/examples/src/keyring/raw_rsa/keypair.py#L38-L46

Issue #139, opened 2020-06-18
Determine if provider info should be part of the KMS call on Decrypt in the AWS KMS Keyring


acioc:
Should we say that provider info SHOULD/MUST be provided to the AWS KMS decrypt call? Do we do this for all AWS KMS decrypt calls?

mattsb42-aws:
Arguments for MUST:

- It adds service-side checks to verify that the CMK is the one that we think it is.
    - There have previously been discussions about whether the AWS KMS keyrings MUST check the CMK ARN in the response on decrypt to make sure that it matches the CMK ARN in the EDK's provider info. If we include this in the KMS call, we can make this a requirement rather than a after-the-fact check. This is very much in line with "correct by construction" and the removal of keyring trace.
- It paves the way for an AWS KMS keyring that always writes the key name as the provider info rather than writing the CMK ARN in the kms:Encrypt/GenerateDataKey response.
    - This would enable using single-CMK keyrings for decryption that use CMK identifiers other than CMK ARN.
    - If all AWS KMS CMKs specify the provider info on kms:Decrypt calls, then the difference between "AWS KMS keyring that writes the CMK ARN in the response as provider info" and "AWS KMS keyring that writes the key name as provider info" *could* collapse to a simple optional configuration flag on initialization.

acioc:
Closing in favor of https://github.com/awslabs/aws-encryption-sdk-specification/issues/165

Issue #94, opened 2020-04-24
define client supplier interface and common manifestations


mattsb42-aws:
Related to #15, #16, and #90: we have never really defined anything about client suppliers beyond "they supply clients"[1]. This has resulted in a lot of churn and cases of implementations leading the spec rather than implementing the spec.

We need to at least define:
- [ ] What is a client supplier's role/purpose?
- [ ] What is the client supplier interface?
- [ ] What client suppliers are part of the "standard framework" and what are their interfaces and behaviors?

[1] https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md#client-supplier

mattsb42-aws:
One point we need to make sure we hit in here is: what does a caching client supplier do and why.

acioc:
Closing in favor of https://github.com/awslabs/aws-encryption-sdk-specification/issues/165

Issue #90, opened 2020-04-17
ExcludeRegions/LimitRegions Client Supplier Null/Empty Region Specification


acioc:
We should have a clear specification around the behavior of the ExcludeRegions/ LimitRegions client suppliers and ensure this behavior is matched by all ESDK's. Although most/all ESDKs interpret the definitions of these client suppliers in similar ways, the behavior of handling null/empty regions could be different.

Specifically..
1. How does the LimitRegions client supplier handle null/ no regions? If no region is passed in, should the client supplier GetClient fail? Proposal: Yes, it should fail, since there's no way to know if the default region (which will be host dependent) is part of the limited regions to get clients for.
2. How does the ExcludeRegions client supplier handle null/ no regions? If no region is passed in, should the client supplier GetClient fail? Proposal: Yes, it should fail, since there's no way to know if the default region (which will be host dependent) is part of the regions to avoid.

I believe that, currently, some ESDKs just check `region in regions` or `!(region in regions)` but this behavior (and the handling of nulls) could be language specific.

seebees:
The question is does this happen before or after the call to the underlying supplier.  If it happens _before_ then I assert that you should put in whatever value will be used to denote "Let the SDK work out the region"

mattsb42-aws:
The way I interpret the two is:

* AllowRegion
  * This client supplier MUST accept one or more region name value.
  * This client supplier MUST block any requests for values not in that list.
* DenyRegion
  * This client supplier MUST accept one or more region name values.
  * This client supplier MUST block any requests for values in that list.

Carrying on from that, as @seebees stated, one valid value for "region name value" is "whatever indicates to a client supplier that the desired region is unknown".

lavaleri:
I feel like letting the user specify "idk" in the region list could be very easy to misunderstand if not documented properly. (and maybe even if documented properly...)

e.g.

For excludeRegionsClientSupplier, it would be very easy for a user to make the assumption that, if region A is in that list of excluded regions, that means I can assume that this client supplier will never pass me a client from region A. In order to make these assumptions, a user will *have* to make sure they configure that list with "idk".

For a limitRegionsClientSupplier, it would be very easy for a user to make the assumption that, because region A is not in that list of allowed regions, that means I can assume that this client supplier will never pass me a client from region A. In order to make any such assumption, they will need to make sure "idk" is *not* in that list.

[...]

Issue #84, opened 2020-03-25
Require users to describe intent on KMS Keyring constructor


lavaleri:
The KMS Keyring will have drastically different behavior based on whether it is a discovery keyring or not (see https://github.com/awslabs/aws-encryption-sdk-specification/issues/83). The spec describes the conditions for the KMS behaving one way vs. the other, however does not prescribe on KMS Keyring construction that such user intent is taken in as input. This could result in users mistakenly creating one type of keyring when intending on constructing the other.

We should update the spec to require that implementations take in some isDiscovery param in the construtcor (or something isomorphically similar) such that construction fails if the input is invalid for the type of keyring the user intends to create.

acioc:
This is blocked due to ongoing client supplier/ KMS keyring work

acioc:
Closing in favor of https://github.com/awslabs/aws-encryption-sdk-specification/issues/165

PR #148, opened 2020-07-02
feat: Proposal for re-designing the AWS KMS keyring

acioc:
_Issue #, if available:_

This proposal affects the following issues: https://github.com/awslabs/aws-encryption-sdk-specification/issues/139, https://github.com/awslabs/aws-encryption-sdk-specification/issues/94, https://github.com/awslabs/aws-encryption-sdk-specification/issues/90, https://github.com/awslabs/aws-encryption-sdk-specification/issues/84, https://github.com/awslabs/aws-encryption-sdk-specification/issues/59, https://github.com/awslabs/aws-encryption-sdk-specification/issues/16, https://github.com/awslabs/aws-encryption-sdk-specification/issues/15, https://github.com/awslabs/aws-encryption-sdk-specification/issues/49, https://github.com/awslabs/aws-encryption-sdk-specification/issues/83, https://github.com/awslabs/aws-encryption-sdk-specification/issues/40

_Description of changes:_

The AWS Encryption SDK’s [AWS KMS Keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md) has become increasingly complex as it supports various customer use cases. In this proposal, we separate the functionality of the AWS KMS keyring into multiple smaller-scoped keyrings. Each new keyring maintains a smaller API surface area.

By simplifying the AWS KMS keyring, we establish the basic building blocks for interacting with AWS KMS. We can support additional customer use cases by writing operations to configure a multi-keyring with these smaller-scoped keyrings. In the future, new features will be supported without refactoring code or updating customer-facing APIs. Advanced customers can also follow a similar process for more complex customization. This lowers the likelihood that advanced customers fork ESDK code.

### DIFF
--- a/proposals/2020-07-01_aws-kms-keyring-redesign/background.md
+++ b/proposals/2020-07-01_aws-kms-keyring-redesign/background.md
@@ -244,15 +244,6 @@ The ESDK MUST provide customers a way to use the following:
       2. with each AWS KMS symmetric region discovery keyring configured with the configured AWS account ID, if applicable
       3. with each AWS KMS symmetric region discovery keyring configured with the configured grant tokens, if applicable
       4. where the order of the child keyrings matches the order of the configured AWS regions
-3. **AWS KMS symmetric all region discovery keyring**
-   1. MAY be configured with an AWS account ID
-   2. MAY be configured with a list of string [grant tokens](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#grant-tokens)
-      to be included in all AWS KMS calls
-   3. MUST configure an AWS KMS symmetric multi-region discovery keyring...
-      1. with a list of all AWS regions
-         1. obtained directly from the AWS SDK or from a configuration provided by the AWS SDK
-      2. with the configured AWS account ID, if applicable
-      3. with the configured grant tokens, if applicable
 
 **All keyring-producing operations MUST...**
 
@@ -648,13 +639,14 @@ The AWS KMS discovery keyring limits communication to the single AWS region it i
 
 ### Issue 8: Be able to attempt decryption without needing to specify the CMKs to use for decryption
 
-Customers want to maintain the existing _Is Discovery_ keyring use case.
 See [_Issue 8a_](#issue-8a-optionally-limit-and-order-the-cmks-to-attempt-decryption-with-based-on-their-aws-region-or-aws-account-id).
 
 ### Issue 8a: Optionally limit and order the CMKs to attempt decryption with based on their AWS region or AWS account ID
 
 Customers want to maintain the existing [AWS KMS regional discovery keyring](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/choose-keyring.html#kms-keyring-regional)
-use case.
+use case,
+where a keyring can _attempt_ decryption for any encrypted data key in a specific region,
+regardless of the AWS KMS CMK used to encrypt it.
 We also want to enable customers to have ordered region limiting,
 where customers provide an ordered list of AWS regions
 and decryption is _attempted_ in the provided order.
@@ -671,9 +663,14 @@ and configures it with the optional AWS account ID.
 Each AWS KMS discovery keyring is a child in a multi-keyring,
 in the same order as the regions configured by the customer.**
 
-**A keyring-producing operation that is only configured with an optional AWS account ID also exists.
-This operation initializes AWS KMS discovery keyrings in all AWS regions.
-Each AWS KMS discovery keyring is a child in a multi-keyring.**
+**If customers need a keyring that _attempts_ decryption in all AWS regions,
+they SHOULD call a service/API
+to get an updated list of AWS regions.
+This will prevent any ESDK or AWS SDK region-list from becoming stale over time.
+In most cases, customers SHOULD simply use the keyring-producing operation
+for the specific regions they need.
+This will provide flexibility for adding more regions over time,
+without allowing access to regions that are not currently required.**
 
 1. By configuring the keyring-producing operation,
    customers configure the region(s) and AWS account ID they want to communicate with.
@@ -683,7 +680,6 @@ Each AWS KMS discovery keyring is a child in a multi-keyring.**
    1. Multi-keyring behavior is unchanged
    2. Supports new ordered region filtering use case
    3. Supports new AWS account ID filtering use case
-   4. Customers can access a keyring that will _attempt_ decryption in all AWS regions
 
 ```
 // Without a keyring-producing operation, the proposal is represented as
@@ -704,23 +700,6 @@ multiKeyring = MakeAwsKmsSymmetricMultiRegionDiscoveryKeyring(["us-west-2", "us-
 
 // If customers do not need to limit the AWS account ID...
 multiKeyring = MakeAwsKmsSymmetricMultiRegionDiscoveryKeyring(["us-west-2", "us-east-1", ...])
-
-// If customers do not want to configure any regions,
-// a keyring-producing operation can produce
-// an AWS KMS symmetric all region discovery keyring.
-//
-// The keyring-producing operation uses the AWS SDK
-// to determine all AWS regions
-// and configures one AWS KMS discovery keyring per AWS region.
-//
-// This would be the same as using a multi-region discovery keyring
-// configured with all AWS regions:
-// MakeAwsKmsSymmetricMultiRegionDiscoveryKeyring([every AWS region in list])
-anyRegionDiscoveryKeyring = MakeAwsKmsSymmetricAllRegionDiscoveryKeyring()
-
-// Customers can configure a keyring that limits AWS KMS calls
-// to a specific AWS account ID, but MAY communicate with all AWS regions.
-myDiscoveryKeyring = new MakeAwsKmsSymmetricDiscoveryKeyring(awsAccountId)
 ```
 
 ### Issue 8b: Ensure it is possible to efficiently limit decryption to specific CMKs, even if the decryption operation has a large number of CMKs that my credentials have access to
--- a/proposals/2020-07-01_aws-kms-keyring-redesign/background.md
+++ b/proposals/2020-07-01_aws-kms-keyring-redesign/background.md
@@ -23,11 +23,11 @@ through a derived [_Is Discovery_](https://github.com/awslabs/aws-encryption-sdk
 
 The _AWS KMS discovery keyring_ changes the underlying behavior of the AWS KMS keyring.
 It returns the unchanged encryption materials on encryption and _attempts_ to decrypt any encrypted data key,
-as long as it can create/initialize an AWS SDK KMS service client (AWS SDK client).
+as long as it can initialize an AWS SDK KMS service client (AWS SDK client).
 
 To communicate with AWS KMS, the ESDK’s AWS KMS keyring needs a way to initialize AWS SDK clients.
 Current implementations use a [client supplier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#client-supplier)
-that initializes/configures AWS SDK clients.
+that initializes AWS SDK clients.
 AWS SDK clients are supplied to the AWS KMS keyring at the keyring’s API runtime.
 Client suppliers can fail to supply clients, resulting in API runtime failures.
 Client suppliers can also maintain additional logic,
@@ -105,7 +105,7 @@ with each keyring maintaining a relatively small surface area.
 To handle multiple strings identifying AWS KMS CMKs
 ([key names](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names))
 or multiple AWS regions,
-the proposal uses keyring-producing operations
+the proposal uses what the document calls _keyring-producing operations_
 that output a multi-keyring of smaller-scoped child AWS KMS keyrings.
 Each child AWS KMS keyring is tied to a single AWS SDK KMS service client (AWS SDK client)
 that is initialized before child keyring initialization.
@@ -115,7 +115,7 @@ the basic building blocks for long-term maintainability and extensibility are es
 ESDK developers can support additional customer use cases by writing new keyring-producing operations
 without needing to refactor existing code or update customer-facing APIs.
 Customers are given the same building blocks,
-which allow for additional customization and reduce the need for forked code.
+which allows for additional customization and reduces the need for forked code.
 The smaller components reinforce an "easy to use, hard to misuse" mindset.
 
 ### AWS KMS symmetric keyring
@@ -124,42 +124,39 @@ For the purpose of this document,
 the proposed design’s _AWS KMS symmetric keyring_ will be referred to as the _AWS KMS keyring_
 for brevity.
 
+**Initialization**
+
+On keyring initialization, an AWS KMS symmetric keyring MUST define the following:
+
+- [Key Name](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names)
+- AWS SDK KMS service client (AWS SDK client)
+
+On keyring initialization, an AWS KMS symmetric keyring MAY define the following:
+
+- Grant Tokens
+
 **Behavior**
 
 The AWS KMS symmetric keyring MUST...
 
-1. Be configured with a single string identifying an AWS KMS CMK
-   ([key name](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names))
-   and MUST be configured with a single AWS SDK KMS service client (AWS SDK client) for that key name.
-2. Use the key name and AWS SDK client provided at initialization for encryption and decryption.
-3. Fail on decrypt if the encrypted data key’s provider info does not match the keyring’s configured key name.
-4. Provide the encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
+1. Be configured with a single key name
+   and a single AWS SDK client for that key name.
+   1. The AWS SDK client MUST be configured with credentials for an AWS principal
+      that passes an [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html)
+      authorization check for the encrypted data key.
+2. Use the AWS SDK client provided at initialization for all AWS KMS API calls.
+3. Have an AWS KMS symmetric decryption contract
+   that ensures the encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
+   matches the keyring’s configured key name.
+4. Have an AWS KMS symmetric decryption contract
+   that ensures the encrypted data key’s encryption context exactly matches the encryption context used on encrypt.
+5. Provide the encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
    as part of the [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html) call.
 
 The AWS KMS symmetric keyring MAY...
 
 1. Be configured with a list of string [grant tokens](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#grant-tokens)
-   to be included in all AWS KMS calls.
-
-**Decryption Contract**
-
-1. The keyring MUST have access to an AWS KMS client that is configured with credentials for an AWS principal
-   that passes an [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html)
-   authorization check for the encrypted data key.
-2. The encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
-   MUST match the keyring’s configured key name.
-3. The encryption context MUST exactly match the encryption context used on encrypt.
-
-**Initialization**
-
-On keyring initialization, an AWS KMS symmetric keyring MUST define the following:
-
-- Key Name
-- AWS SDK client
-
-On keyring initialization, an AWS KMS symmetric keyring MAY define the following:
-
-- Grant Tokens
+   to be included in all AWS KMS API calls.
 
 ### AWS KMS symmetric region discovery keyring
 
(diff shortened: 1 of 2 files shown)


proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+**Decryption Contract**
+
+1. The keyring MUST have access to an AWS KMS client that is configured with credentials for an AWS principal
+   that passes an [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html)
+   authorization check for the encrypted data key.
+2. The encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
+   MUST match the keyring’s configured key name.
+3. The encryption context MUST exactly match the encryption context used on encrypt.
+
+**Initialization**
```
robin-aws, line 211: This section seems entirely redundant w.r.t. the Behaviour section. I'd move the more detailed initialization description into this section. Consider also moving Initialization before Behaviour.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+- Key Name
+- AWS SDK client
+
+On keyring initialization, an AWS KMS symmetric keyring MAY define the following:
+
+- Grant Tokens
+
+### AWS KMS symmetric region discovery keyring
+
+Customers MUST declare their intent to use the AWS KMS symmetric region discovery keyring instead of the AWS KMS symmetric keyring.
```
robin-aws, line 224: I think we need to be more precise about "declare their intent" here. I think what we want is "this MUST be a different keyring" but that's a fuzzy statement. Perhaps just "the keyring-producing operations that produce this kind of keyring MUST be distinct from those that produce non-discovery keyrings"?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+
+1. Be configured with a single AWS SDK client.
+   1. Determine the AWS region associated with the provided AWS SDK client.
+      In some implementations,
+      this MAY require an additional AWS region argument
+      during AWS KMS symmetric region discovery keyring initialization.
+      If the AWS SDK returns the AWS SDK client’s region
+      directly from the AWS SDK client
+      in an intuitive manner (by using a single function/method call),
+      implementations MUST NOT require an additional AWS region argument.
```
robin-aws, line 238: This will be tricky to strictly enforce, since it could easily be violated if a newer version of the SDK adds this "intuitive manner" under us. We'd have to set up a canary test case that asserts the exact maximum SDK version we depend on, so that it will break and force us to check on every version bump. It also implies a breaking change if this happens, since we then have to remove the parameter and force customers to stop providing it.

I'm not saying we shouldn't do this, but I'd feel better if we checked and found all current SDKs DO support this. Do we have hard evidence this currently isn't possible in any of them? Should we push for all SDKs to?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+2. Be configured with a list of string [grant tokens](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#grant-tokens)
+   to be included in all AWS KMS calls.
+
+**Decryption Contract**
+
+1. The keyring MUST have access to an AWS KMS client that is configured with credentials for an AWS principal
+   that passes an [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html)
+   authorization check for the encrypted data key.
+2. The keyring is unable to write a decryption contract,
+   but it MAY fulfill any AWS KMS decryption contract for the configured AWS region and, if provided, the AWS account ID.
```
robin-aws, line 262: I don't think "AWS KMS decryption contract" is ever really defined. It should at least be "AWS KMS symmetric decryption contract". In fact, would it be more accurate and simpler to say "any decryption contract written by an AWS KMS symmetric keyring"?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+      1. with a single AWS KMS symmetric region discovery keyring for each configured AWS region
+      2. with each AWS KMS symmetric region discovery keyring configured with the configured AWS account ID, if applicable
+      3. with each AWS KMS symmetric region discovery keyring configured with the configured grant tokens, if applicable
+      4. where the order of the child keyrings matches the order of the configured AWS regions
+
+**All keyring-producing operations MUST...**
+
+1. Initialize/configure the AWS SDK client(s) required for communicating with AWS KMS for the given key names/AWS regions
+2. Configure each AWS KMS keyring with the AWS SDK client it needs to communicate with AWS KMS
+3. Be configurable with an optional AWS SDK client configuration (client config)
```
robin-aws, line 306: Can you link to an example here?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+**All keyring-producing operations MAY...**
+
+1. Add an identifier to the user agent string in the AWS SDK client
+   that was initialized by the keyring-producing operation
+   to note the ESDK language and version
+
+**All keyring-producing operations MUST NOT...**
+
+1. Modify any customer-provided client config
+2. Modify the user agent string if the customer has already provided one
```
robin-aws, line 326: The user agent string is generally a concatenation of separate strings, and the SDKs generally make this an additive process (e.g. https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/RequestClientOptions.html#appendUserAgent-java.lang.String-). Assuming the more important requirement of not modifying parameters, do we need this restriction?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+On keyring initialization, an AWS KMS symmetric region discovery keyring MUST define the following:
+
+- AWS SDK client (and the AWS SDK client’s region)
+
+On keyring initialization, an AWS KMS symmetric region discovery keyring MAY define the following:
+
+- AWS account ID
+- Grant Tokens
+
+### Keyring-Producing Operations
```
robin-aws, line 276: I'm still having trouble with terminology here: we're describing these half as types of keyring with specific configurations and half as an operations that produce keyrings. 

I think we should separate the two things better. It's hard to pick out the right terminology for "kinds" or "types" of keyrings since it's completely implicit in the current specification. I'd suggest we call these "Derived Keyrings" instead, so we're describing the keyrings rather than the operations.

The rest of my comments assume this change.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+
+**Please note the following names are not finalized.**
+
+The ESDK MUST provide customers a way to use the following:
```
robin-aws, line 280:
```suggestion
The ESDK MUST provide customers a way to initialize the following:
```

Or you could say "MUST provide keyring-producing operations for the following". The concept of "keyring-producing operation" is something we should introduce at a higher level to clarify how that relates to all the Initialization sections.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+   4. MUST configure a multi-keyring of AWS KMS symmetric keyrings...
+      1. with a single AWS KMS symmetric keyring for each configured key name
+      2. with each AWS KMS symmetric keyring configured with the configured grant tokens, if applicable
+      3. where the order of the child keyrings matches the order of the configured key names
+2. **AWS KMS symmetric multi-region discovery keyring**
+   1. MUST be configured with a list of AWS regions
+   2. MAY be configured with an AWS account ID
+   3. MAY be configured with a list of string [grant tokens](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#grant-tokens)
+      to be included in all AWS KMS calls
+   4. MUST configure a multi-keyring of AWS KMS symmetric region discovery keyrings...
```
robin-aws, line 296:
```suggestion
   4. MUST be implemented as a multi-keyring of AWS KMS symmetric region discovery keyrings...
```

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+      1. with a single AWS KMS symmetric region discovery keyring for each configured AWS region
+      2. with each AWS KMS symmetric region discovery keyring configured with the configured AWS account ID, if applicable
+      3. with each AWS KMS symmetric region discovery keyring configured with the configured grant tokens, if applicable
+      4. where the order of the child keyrings matches the order of the configured AWS regions
+
+**All keyring-producing operations MUST...**
```
robin-aws, line 302: Consider making this something like "The ESDK MUST provide keyring-producing operations for these derived keyrings. All such operations MUST..." Here you really are talking about the operations rather than the end result. All of these last sections should be qualified to refer only to these derived cases, assuming we define "keyring-producing operations" to also apply to initializing the keyring base cases.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+3. [AWS KMS asymmetric CMKs](https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-concepts.html#asymmetric-cmks)
+
+## Motivation
+
+1. **Move logic out of the AWS KMS keyring** so a single keyring does not handle
+   multiple AWS KMS [CMKs](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#master_keys)
+   ([key names](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names)),
+   multiple AWS regions,
+   or complex functionality (
+   [discovery](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#is-discovery),
```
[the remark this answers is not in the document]
acioc: Yup! This still interprets properly! Updating here and in background.md. 

For this change, and any suggestions like this below, I'll update them myself since "Add suggestion to batch" doesn't appear to work. Also, in some cases, I have to update both documents so doing it myself ensures I don't forget any changes.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+   This lowers the potential for keyring API runtime failures.
+   There is clear intent/knowledge of _which_ keyring is using _which_ AWS SDK client
+   since keyrings are explicitly configured with an AWS SDK client.
+   This makes AWS SDK clients easy to use, hard to misuse.
+4. **Support additional customer use cases** that cannot be supported today without significant refactors.
+   For example, we now support region and key name ordering when _attempting_ decryption.
+   For new supported use cases, see [Background](background.md).
+   We can continue developing the ESDK with an “easy to use, hard to misuse” mindset.
+
+## Drawbacks
```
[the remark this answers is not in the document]
acioc: Added note on documentation and usage examples.

Also added note to see Background for pseudocode.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+since any shared clients MUST be configured at keyring initialization instead of keyring API runtime.
+
+## Operational Implications
+
+We MUST re-implement the AWS KMS keyring in all languages the ESDK supports.
+Existing ESDKs that have released AWS KMS keyrings require a migration/deprecation path.
+
+The [ESDK for C](https://github.com/aws/aws-encryption-sdk-c/blob/aa85ca224d550cfe110e2112821a84506b9aca3e/aws-encryption-sdk-cpp/source/kms_keyring.cpp)
+implements the AWS KMS keyring in C++ and returns generic keyring structs.
+The new AWS KMS keyring implementations can be added and the existing keyring implementations can be deprecated.
```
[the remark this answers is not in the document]
acioc: Sure! I added a note. Here we can just use AWS KMS as well, since it looks like that naming was never updated.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+Current implementations use a [client supplier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#client-supplier)
+that initializes/configures AWS SDK clients.
+AWS SDK clients are supplied to the AWS KMS keyring at the keyring’s API runtime.
+Client suppliers maintain logic for limiting communication to specific AWS regions and caching AWS SDK clients.
+
+To reduce the scope of the AWS KMS keyring,
+we are moving the keyring’s functionality into multiple smaller-scoped keyrings.
+These smaller-scoped keyrings are configured by what the document calls _keyring-producing operations_.
+Depending on the implementation, keyring-producing operations can be thought of as factory methods/builders.
+They initialize/configure keyrings.
```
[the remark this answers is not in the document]
acioc: I will work on the wording a bit more - at the very least, I won't have a `/` here. With that said, this kind of depends on the interpretation. If you're implementing this as a builder, you configure a builder with a set of parameters, and then the builder initializes a thing with those parameters (or at least that's how I view it) - you configure the parameters you use the initialize something.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+logic is broken up into multiple smaller-scoped keyrings.
+Rather than support multiple keys/AWS regions in a single AWS KMS keyring,
+we use keyring-producing operations to configure a [multi-keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/multi-keyring.md)
+of these smaller-scoped keyrings.
+The smaller-scoped AWS KMS keyrings become the basic building blocks that keyring-producing operations use to satisfy different customer use cases.
+
+From a customer standpoint, naming will be clearer.
+Customers interact with the keyring-producing operations and do not directly interact with the basic building blocks.
+However, more advanced customers can directly interact with the smaller-scoped keyrings
+and write their own keyring-producing operations.
```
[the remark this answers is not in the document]
acioc: That's fair! Updating.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,327 @@
...
+
+**Behavior**
+
+The AWS KMS symmetric keyring MUST...
+
+1. Be configured with a single string identifying an AWS KMS CMK
+   ([key name](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names))
+   and MUST be configured with a single AWS SDK KMS service client (AWS SDK client) for that key name.
+2. Use the key name and AWS SDK client provided at initialization for encryption and decryption.
+3. Fail on decrypt if the encrypted data key’s provider info does not match the keyring’s configured key name.
```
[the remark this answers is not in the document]
acioc: Yup. Removing. It's a match for:

"The encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information) MUST match the keyring’s configured key name."

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Moving initialization before behavior.

I don't believe (most of) the information is redundant, but I agree having two separate sections is redundant.  Will combine.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: I'm leaving this sentence as-is, but adding an additional sentence for further clarification.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: My understanding is that the Java AWS SDK makes it difficult to obtain the AWS region *after* the AWS SDK client has been initialized.

This wouldn't necessarily be a breaking change. From an implementation standpoint this would likely be...
```
Before:
MakeKeyring(sdkClient, region):
  // Do stuff


After:
MakeKeyring(sdkClient):
  region = GetRegion(sdkClient);
  MakeKeyring(sdkClient, region);

MakeKeyring(sdkClient, region):
  // Do Stuff
```

`MUST NOT require` Does not imply `MAY not provide`, so supporting both is still upholding this behavior.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Fair point. Updating to add clarity.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Added...

   (examples include: [Java ClientConfiguration](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/section-client-configuration.html), [Python botocore.config](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html), etc.)

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Leaving this open. We'll need to have more discussions about this. I do not believe we need to be this restrictive, but this decision may be outside the scope of this PR.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Leaving this open for now, will address as part of a separate commit (in this PR).

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: I already introduce the concept higher up, but will change to initialize here (though I will most likely rephrase this for the "derived keyring" change (or other naming change).

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Updating.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: I updated the text above the derived keyrings to read "The ESDK MUST provide keyring-producing operations for the following derived keyrings:"

I think leaving "**All keyring-producing operations MUST...**" as-is works best. We...
1. Define what keyring-producing operations are/do earlier in the doc
2. Say which ones we need
3. Say "All keyring-producing operations MUST/MAY/MUST NOT"

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Added term for "derived keyring". I also moved sections around to (hopefully) make things clearer.

proposals/2020-07-01_aws-kms-keyring-redesign/background.md
```diff
@@ -0,0 +1,1396 @@
...
+The AWS KMS symmetric keyring MUST...
+
+1. Be configured with a single key name
+   and a single AWS SDK client for that key name.
+   1. The AWS SDK client MUST be configured with credentials for an AWS principal
+      that passes an [AWS KMS Decrypt API](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html)
+      authorization check for the encrypted data key.
+2. Use the AWS SDK client provided at initialization for all AWS KMS API calls.
+3. Have an AWS KMS symmetric decryption contract
+   that ensures the encrypted data key’s [provider info](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/structures.md#key-provider-information)
```
seebees, line 150: This means that on decrypt the provided EDK's provider info MUST match
the configured value.
As opposed to on encrypt the created EDK's provider info should match the configured value

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,340 @@
...
+   limiting communication to specific AWS regions,
+   AWS SDK client initialization, etc.).
+   The smaller scope of the new keyrings reduces the risk of bugs during both ESDK development and ESDK usage.
+   This makes them easy to use, hard to misuse.
+2. **Create the basic building blocks for interacting with AWS KMS**
+   by creating an AWS KMS symmetric keyring and a separate AWS KMS symmetric region discovery keyring
+   (see [Reference-level Explanation](#reference-level-explanation)).
+   New features are supported by writing operations that configure one or more of these smaller-scoped keyrings.
+   These operations also serve as examples, to show customers how we use the basic building blocks for more complex use cases.
+   This lowers the risk for one-way doors and makes the customer-facing API easy to use, hard to misuse.
```
seebees, line 89: It also forces us to use our own APIs.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,340 @@
...
+They initialize keyrings.
+
+Keyring-producing operations act as the customer-facing APIs.
+Like client suppliers, they initialize AWS SDK clients.
+However, AWS SDK clients are now initialized before a keyring is configured.
+Keyring-producing operations also handle the load of manually initializing multiple, smaller-scoped keyrings.
+They initialize all keyrings required to meet a specific customer use case.
+We refer to the keyrings that keyring-producing operations produce as _derived keyrings_.
+If a use case involves multiple AWS regions or multiple key names,
+the keyring-producing operation will initialize a [multi-keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/multi-keyring.md)
```
seebees, line 171: I think it should _always_ be a multi-keyring.
Even if there is only 1 thing involved.
These keyring-producing operations MUST identify the generator,
this makes a single multi-keyring different
then a single AWS KMS keyring.
And can cause surprises.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,340 @@
...
+and write their own keyring-producing operations.
+This gives advanced customers more flexibility and reduces the need to fork ESDK code in order to meet a specific use case.
+
+## Reference-level Explanation
+
+### AWS KMS symmetric keyring
+
+**Initialization**
+
+On keyring initialization, an AWS KMS symmetric keyring MUST define the following:
```
seebees, line 194: would "accept" be better than "define" in this case?

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
```diff
@@ -0,0 +1,340 @@
...
+
+- [Key Name](https://github.com/awslabs/aws-encryption-sdk-specification/blob/dbc17f93100667e28dc54e64d05a625db3e5bac2/framework/kms-keyring.md#key-names)
+- AWS SDK KMS service client (AWS SDK client)
+
+On keyring initialization, an AWS KMS symmetric keyring MAY define the following:
```
seebees, line 199: Should this MAY stand?
For me the issue is "MAY define",
if it was "MAY receive"...

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: I agree - I'll rephrase to capture this better.

proposals/2020-07-01_aws-kms-keyring-redesign/background.md
acioc: Yup! I'll add clarification. Will likely copy/paste the existing spec language from https://github.com/awslabs/aws-encryption-sdk-specification/blob/be6870b2513f4ca44bea8f2e0b5eb4808dba4365/framework/kms-keyring.md#key-names

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Adding note.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Will rephrase based on the above.

proposals/2020-07-01_aws-kms-keyring-redesign/change.md
acioc: Will rephrase. This is in-line with the existing AWS KMS keyring language, but I noticed other keyrings (AES and RSA) say "inputs are REQUIRED" so it looks like we don't actually have a standard for this :(

Will create an issue to standardize this language

acioc: Created https://github.com/awslabs/aws-encryption-sdk-specification/issues/151

proposals/2020-07-01_aws-kms-keyring-redesign/background.md
```diff
@@ -194,83 +211,77 @@ The AWS KMS symmetric region discovery keyring MAY...
...
+      2. Keyrings MUST communicate with AWS KMS in the same AWS region
+      3. AWS SDK clients MUST share the same AWS SDK client configuration
+6. Maintain the order of all key names/AWS regions provided by the customer
 
-On keyring initialization, an AWS KMS symmetric region discovery keyring MAY define the following:
+**All keyring-producing operations MAY...**
 
-- AWS account ID
-- Grant Tokens
+1. Add an identifier to the user agent string in the AWS SDK client
```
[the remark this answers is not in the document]
acioc: This only applies to the keyring-producing operations for now, since that's where we construct the AWS SDK client. I believe in most languages, the client is directly associated with a Config that contains the user agent string, so we don't update this again once the client is created.

During the implementation phase, we'll determine if there are languages were the user agent string is instead passed per call instead of per client. If the string is passed when the call is made and cannot be passed directly to the client for all calls, we might need to amend this behavior a bit.

proposals/2020-07-01_aws-kms-keyring-redesign/background.md
```diff
@@ -194,83 +211,77 @@ The AWS KMS symmetric region discovery keyring MAY...
...
+   that was initialized by the keyring-producing operation to note the ESDK language and version
 
-### Keyring-Producing Operations
+**All keyring-producing operations MUST NOT...**
 
-Depending on the implementation, keyring-producing operations can be thought of as factory methods/builders.
-They initialize/configure keyrings.
+1. Modify any customer-provided client config
+2. Modify the user agent string if the customer has already provided one
```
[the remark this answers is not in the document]
acioc: I'll try to clarify this a bit more.

Issue #136, opened 2020-06-16
Add implementation guidance on when to decrypt data


mattsb42-aws:
One of the promises that the ESDK makes is that it will not release plaintext until it has been authenticated. For framed messages this means authenticating using the GCM tag for a frame, while for unframed messages this means authenticating using the GCM tag for the message body.

We should add specific implementation guidance on the order of operations to follow. Specifically:

1. Read entire ciphertext (one frame or the entire unframed message body) into a ciphertext buffer.
1. Decrypt the ciphertext into a plaintext buffer and authenticate using the tag.
1. Release the plaintext to the caller.

mattsb42-aws:
background:

GCM is just CTR+GMAC, and you have to completely process the ciphertext before you can authenticate the GMAC tag. The way most GCM implementations work is that they stream the plaintext out to you as you go through decrypting the ciphertext, then when you think you’re done you can authenticate the message and AAD against the tag. This pattern is necessary for low-resource-overhead streams but is problematic because it means that you are releasing the plaintext to the caller before you have authenticated it, thereby losing a lot of the value that GCM provides over CTR. To mitigate this issue, we add this buffering layer on top of the underlying GCM implementation, and only start releasing plaintext once we have authenticated the entire ciphertext.

robin-aws:
See also #142 - if the caller provides an encryption context it should also be verified before releasing any plaintext.

mattsb42-aws:
@robin-aws I think that is best handled under a different scope. The scope of this specific issue is meant to be for the component that is handling the message body. By definition, by the time the client reaches the body, the encryption context is no longer in play. We have a few different levers we can pull on the encryption context (keyring, header auth), and we should perhaps make more specific statements about what MUST and SHOULD happen at each of those stages, but I do not think that #142 is relevant to this specific scope.

robin-aws:
Okay fair enough - so my statement isn't incorrect, it's just not relevant here.

lavaleri:
This is specified in #149

Issue #27, opened 2019-08-22
Add streaming decrypt spec


ttjsu-aws:
Update Decrypt API to include specification of operation in streaming context (i.e. how should this operate in chunks, not requiring everything in memory?)

Current gap:
- How should verification of the signature (if included) be performed?
- How/when should plaintext be released?
- How/when does the EC become available?
- Various inputs effects' on streaming behavior
- Implications on memory required to be buffered
- Edge cases with possible language specific streaming constructs (i.e. make sure spec covers behavior if a stream is flushed or closed)

Related:
- [ ] https://github.com/awslabs/aws-encryption-sdk-specification/issues/113

lavaleri:
Completed in #149

Issue #26, opened 2019-08-22
Add streaming encrypt spec 


ttjsu-aws:
Update Encrypt API to include specification of operation in streaming context (i.e. how should this operate in chunks, not requiring everything in memory?)

Current gap:
- How should calculation of the signature (if included in alg suite) be performed?
- How/when should ciphertext be released?
- How/when does the EC become available?
- Various inputs effects' on streaming behavior
- Implications on memory required to be buffered
- Edge cases with possible language specific streaming constructs (i.e. make sure spec covers behavior if a stream is flushed or closed)

lavaleri:
Completed in #149

PR #149, opened 2020-07-07
docs: Update Encrypt and Decrypt to clarify streaming

lavaleri:
_Issue #, if available:_ #26 #27 #136

_Description of changes:_ Specify requirements for encrypt/decrypt operating in streaming use case.


client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
-The decrypt behavior MUST then use this plaintext data key, algorithm suite and encryption context, if included, to decrypt the encrypted content
-and obtain the plaintext to be returned. The [encrypted content](../data-format/message-body.md#encrypted-content) to be decrypted is obtained by parsing the
-[message body](../data-format/message-body.md) of the encrypted message inputted.
+The Decrypt operation is divided into several distinct parts:
 
-Note: If the message is framed, the encrypted content to decrypt is stored across multiple frames.  
-If the message is non framed, the encrypted content to decrypt is stored as a single blob.
+- [Parse the header](#parse-the-header)
+- [Get the decryption materials](#get-the-decryption-materials)
+- [Decrypt the body](#decrypt-the-body)
```
SalusaSecondus, line 122: Insert: Verify the header between these two steps.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
-Decrypt MUST use the encryption algorithm obtained from the algorithm suite.  
-The cipher key used for decryption is the derived key outputted by the [KDF algorithm](../framework/algorithm-suites.md#key-derivation-algorithm)
-specified by the algorithm suite.  
-The input to the KDF algorithm is the plaintext data key.
+This operation MUST perform these steps in order.
 
-The AAD used in decryption is the [Message Body AAD](../data-format/message-body-aad.md), constructed as follows:
+If the caller is streaming the encrypted plaintext to this operation
+and indicates an end to the encrypted message such that this operation can
+not complete the above steps,
```
SalusaSecondus, line 129: "cannot" is one word

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+
+This operation MUST wait if it doesn't have enough encrypted message bytes available to
+deserialize the next field of the message header until enough input bytes become available or
+the caller indicates an end to the encrypted message.
 
-[TODO]
+Once a valid message header is deserialized,
+this operation MUST validate the [message header body](../data-format/message-header.md#header-body)
+by using the [authenticated encryption algorithm](../framework/algorithm-suites.md#encryption-algorithm)
+to decrypt with the following inputs:
```
SalusaSecondus, line 153: Validation of the header cannot happen before decryption materials are retrieved.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
 
-## Apendix
+- the AAD is the serialized [message header body](../data-format/message-header.md#header-body).
+- the IV has a value of 0
+- the cipherkey is the derived data key
+- the ciphertext is an empty byte array
```
SalusaSecondus, line 158: Maybe a bigger change, but this definition may better belong in the algorithm suite section rather than here. While all current suites use this technique, future ones may not.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+
+If this decryption fails, this operation MUST immediately halt and fail.
```
SalusaSecondus, line 160: decryption -> tag verification

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+
+If the algorithm suite contains a signature algorithm and
+this operation is streaming the plaintext output to the caller,
+this operation MUST input the serialized frame to the signature algorithm as soon as it is deserialized,
+such that the serialized frame isn't required to remain in memory to complete
```
SalusaSecondus, line 165: Strike this memory comment. Implementations can choose to do less efficient things. Some verification APIs do not support a streaming mode and thus will require it to be held in memory.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+the [signature calculation](#signature-calculation).
+
+If this decryption succeeds and the algorithm suite does not contain a signature algorithm,
+the parsed [encryption context](#encryption-context) MAY be released to the caller.
+
+If the algorithm suite does contain a signature algorithm
+the parsed [encryption context](#encryption-context) MUST NOT be released to the caller until the
+[signature is verified](#verify-the-signature-optional).
```
SalusaSecondus, line 173: I thought we determined it could be released, just like we release the plaintext prior to signature verification.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  used in the message body AAD above.
+- The cipherkey is the derived data key
+- The ciphertext is the [encrypted content](../data-format/message-body.md#encrypted-content).
+
+If this decryption fails, this operation MUST immediately halt and fail.
+
+If the algorithm suite contains a signature algorithm and
+this operation is streaming the plaintext output to the caller,
+this operation MUST input the serialized frame to the signature algorithm as soon as it is deserialized,
+such that the serialized frame isn't required to remain in memory to complete
```
SalusaSecondus, line 265: Same note on memory.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+If the encrypted message includes an algorithm suite without a signature algorithm,
+the result plaintext MAY be released.
+
+If the encrypted message includes an algorithm suite with a signature algorithm
+the result plaintext MUST NOT be released until [signature verification succeeds](#verify-the-signature-optional).
+
+If streaming the output plaintext for a message with a signature,
+the result plaintext MAY be released, however the caller MUST NOT
+consider this plaintext verified.
+See [security considerations](#security-considerations) below.
```
SalusaSecondus, line 277: These two paragraphs feel contradictory. Let's clarify it. (I think this comment applies elsewhere.)

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
...
-  specified by the [algorithm suite](../framework/algorithm-suites.md), with the following inputs:
-  - The AAD is the serialized [message body AAD](../data-format/message-body-aad.md)
-  - The IV is the [IV](../data-format/message-body.md#iv) specified for this frame above.
-  - The cipherkey is the derived data key
-  - The plaintext contains part of the input [plaintext](#plaintext) this frame is encrypting.
-- [Authentication Tag](../data-format/message-body.md#authentication-tag): MUST be the authentication tag outputted by the above encryption.
+- The AAD is the serialized [message header body](../data-format/message-header.md#header-body).
+- The IV has a value of 0.
+- The cipherkey is the derived data key
+- The plaintext is an empty byte array
```
SalusaSecondus, line 198: As earlier, we may want to lift this out to the suites section.

client-apis/encrypt.md
```diff
@@ -83,26 +95,62 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
...
+- [Construct the header](#construct-the-header)
+- [Construct the body](#construc-the-body)
+- [(Optional) Construct the signature](#construct-the-signature)
+
+This operation MUST perform these steps in order.
+These steps calculate and serialize the components of the output [encrypted message](#encrypted-message).
+Any data that is not specified within the [message format](../data-format/message.md)
+MUST NOT be added to the output message.
+
+The bytes of the output encrypted message MAY be released to the caller as soon as they are serialized.
```
seebees, line 126: I don't think so...?
The encrypt operation MUST know where the end is.
Especially for framed messages.
I suppose it could be up to the implementation,
but this seems like something that is easy to get wrong.

e.g. Frame Length is 4K, and I send 100 byte chunks
and then send EOF after 10 chunks (1K).
The header can be released, but
if the body header has been sent
it is probably wrong.

Also, the plaintext length bound
is just an estimate,
so I don't think we can rely on it
as a single to send the final frame.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+This operation MUST obtain this set of [decryption materials](../framework/structures.md#decryption-materials),
+by calling [Decrypt Materials](../framework/cmm-interface.md#decrypt-materials) on a [CMM](../framework/cmm-interface.md).
+
+The CMM used MUST be the input CMM, if supplied.
+If a CMM is not supplied as the input, the decrypt operation MUST construct a [default CMM](../framework/default-cmm.md)
+from the [keyring](../framework/keyring-interface.md) inputted.
+
+The call to CMM's [Decrypt Materials](../framework/cmm-interface.md#decrypt-materials) operation
+MUST include as the input the [encryption context](../data-format/message-header.md#aad)
+(if the encryption context is empty, this operation MAY pass no encryption context),
```
seebees, line 193: On the encrypt side we say:

- Encryption Context: If provided, this is the [input encryption context](#encryption-context).
  Otherwise, this is an empty encryption context.

I think we should say the same thing on both sides for the CMM...

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
 Note that if the key derivation algorithm is the [identity KDF](../framework/algorithm-suites.md#identity-kdf),
 then the derived data key is the same as the plaintext data key.
 
-### Message
+The frame length used in the procedures described below is the input [frame length](#frame-size),
+if supplied.
+If a frame length is not specified on input, this operation MUST choose a reasonable value to
+default to.
```
seebees, line 167: I think that all implementations use 4K.
I can get down with this not being in the spec,
but I would like to be cognitive about that ;)

client-apis/encrypt.md
seebees: If we are thinking of moving this to the suites,
(which I like)
should we also move
the equivalent body parts?
The Body AAD et. al.

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
...
+  - The [Body AAD Content](../data-format/message-body-aad.md#body-aad-content) depends on
+    whether the thing being encrypted is a regular frame or final frame.
+    Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
+  - The [sequence number](../data-format/message-body-aad.md#sequence-number) is the sequence
+    number of the frame being encrypted.
+    If this is the first frame sequentially, this value MUST be 1.
+    Otherwise, this value MUST be 1 greater than the value of the sequence number
+    of the previous frame.
+  - The [content length](../data-format/message-body-aad.md#content-length) MUST have a value
+    equal to the length of the plaintext being encrypted.
```
seebees, line 275: I think that "length of the plaintext" in ambiguous in this context.
Since we are talking about a frames this MUST be the  <= frame length.

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
...
+- The IV is the [sequence number](../data-format/message-body-aad.md#sequence-number)
+  used in the message body AAD above.
```
seebees, line 277: As noted above padded with 0 to 12 bytes. (Which is also specified by the algorithm suite...)

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
...
+  constructed as follows:
+  - The [message ID](../data-format/message-body-aad.md#message-id) is the same as the
+    [message ID](../data-frame/message-header.md#message-id) serialized in the header of this message.
+  - The [Body AAD Content](../data-format/message-body-aad.md#body-aad-content) depends on
+    whether the thing being encrypted is a regular frame or final frame.
+    Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
+  - The [sequence number](../data-format/message-body-aad.md#sequence-number) is the sequence
+    number of the frame being encrypted.
+    If this is the first frame sequentially, this value MUST be 1.
+    Otherwise, this value MUST be 1 greater than the value of the sequence number
```
seebees, line 272: Unless it is the final frame.

client-apis/encrypt.md
```diff
@@ -113,68 +161,168 @@ This document refers to the output of the key derivation algorithm as the derive
...
-- [Signature](../data-format/message-footer.md#signature): MUST be the output of the [signature algorithm](../framework/algorithm-suites.md#signature-algorithm)
-  specified by the [algorithm suite](../framework/algorithm-suites.md), with the following input:
-  - the signature key is the [signing key](../framework/structures.md#signing-key) in the [encryption materials](../framework/structures.md#encryption-materials)
-  - the input to sign is the concatenation of the serialization of the [message header](../data-format/message-header.md) and [message body](../data-format/message-body.md)
+Note that the message header and message body MAY have already been inputted during previous steps.
 
-Any data that is not specified within the [message format](../data-format/message.md) MUST NOT be added to the output message.
+This operation MUST then serialize serialized a message footer with the following specifics:
+
+- [Signature](../data-format/message-footer.md#signature): MUST be the output of the calculation above.
```
seebees, line 315: Well, not exactly.  Since the above calculation does not include a length.

client-apis/encrypt.md
lavaleri: I'm not sure I understand your example. If the operation has successfully serialized a frame, it should be able to release that frame.

client-apis/encrypt.md
lavaleri: This is specified more clearly below, but I've brought that up here as well. I'm unsure how to reduce redundancies here in a good way.

client-apis/encrypt.md
lavaleri: Updated.

client-apis/encrypt.md
lavaleri: Added spec for the signature length field.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
-- Message ID: This value is the same as the message ID in the parsed message header.
-- Body AAD Content: This value depends on whether the encrypted content being decrypted is within a [regular frame](../data-format/message-body.md#regular-frame) ,
-  a [final frame](../data-format/message-body.md#final-frame) or is [non framed](../data-format/message-body.md#non-framed-data).
-  Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
-- Sequence Number: This value is the [sequence number](../data-format/message-body.md#sequence-number) of the frame being decrypted, if the message contains framed data.
-  If the message contains non framed data, then this value is 1.
-- Content Length: TODO
+The caller MAY forgoe indicating an end to the encrypted message.
+This operation MAY still succeed as long as it can successfully process the first
+sequential bytes inputted to it as an encrypted message.
```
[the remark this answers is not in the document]
seebees: I would suggest that the rest should be an error.

client-apis/encrypt.md
lavaleri: Note: Update to SHOULD.

We do not state what it means to be "serialized", so there are parts where this wording is confusing.

client-apis/decrypt.md
```diff
@@ -69,65 +91,226 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  - The [message ID](../data-format/message-body-aad.md#message-id) is the same as the
+    [message ID](../data-frame/message-header.md#message-id) deserialized from the header of this message.
+  - The [Body AAD Content](../data-format/message-body-aad.md#body-aad-content) depends on
+    whether the thing being decrypted is a regular frame, final frame, or unframed data.
+    Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
+  - The [sequence number](../data-format/message-body-aad.md#sequence-number) is the sequence
+    number deserialized from the frame being decrypted.
+    If this is unframed-data, this value MUST be 1.
+    If this is framed data and the first frame sequentially, this value MUST be 1.
+    Otherwise, this value MUST be 1 greater than the value of the sequence number
```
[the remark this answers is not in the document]
lavaleri: Added a statement about using the first bytes to determine whether you are deserializing a final frame vs regular frame.

client-apis/encrypt.md
lavaleri: Update: staying with MAY because I'm unsure how to phrase this while not implying that the operation is the necessarily the actor for releasing the bytes. I think we are still ok as far as stating earlier that you SHOULD NOT extend a streaming API if you need to buffer the whole message.

client-apis/encrypt.md
lavaleri: See related comment above.

client-apis/decrypt.md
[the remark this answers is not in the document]
lavaleri: I'm digging into what we should mean by "ignore" here. Consuming and ignoring is bad. *Not* consuming any more to allow something else to consume might be ok, but is very dependent on how the streaming is framed (i.e. operation is reading, not being written to) and I'm not sure if there is a sane way to do it wrt how implementations read from source.

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+There SHOULD be no operational implications for specifying the streaming use case.
+
+## Guide-level Explanation
+
+The following sections describe how this change is updating the specification
+to answer important questions.
+
+### How should the operation process data in parts?
+
+This change rewites these specifications to
```
acioc, line 67: np: `rewites` -> `rewrites`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+better specify how these operations are allowed to process data in parts.
+Specifically, we attempt to be completely language agnostic and frame streaming as a way for
```
acioc, line 69: np: `and frame streaming` -> `and use frame streaming`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+bytes to move sequentially, over time.
+
+With this framing, the things the encrypt and decrypt operations care about are:
+
+- Is the caller allowed to send any more data?
+- What data is currently available to me?
+
+In the context of the encrypt or decrypt operation, we specify the conditions
+for when the operation MAY or MUST perform some processing.
+These conditions include how many input bytes are avaiable,
```
acioc, line 79: np: `avaiable` -> `available`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+whether more bytes MAY be available,
+at what step in the operation you are in,
+and the result of various calculations.
+
+Additionally, we use similar conditions to specify when the output of the operation MAY be released.
+
+This change SHOULD NOT introduce any language which is implementation specific for
+a specific streaming interface.
+Any streaming interface which is capable of moving bytes over time from one place to another,
+and is able to indicate EOF MUST be able to implement this specification as written.
```
acioc, line 89: np: Maybe add `'s around EOF for better formatting?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+the operation obtains plaintext in parts as it processes the message,
+but has no way of knowing whether the message is verified until the end of the operation.
+So what should we do with the unverified plaintext decrypted thus far?
+Because streaming is being used, we MUST NOT buffer the entire plaintext,
+as that would defeat the point of streaming.
+Thus, the plaintext needs to go somewhere other than memory.
+Because this is a client-side library, we MUST give it to the customer.
+
+In the case of any decryption failure,
+the operation MUST zero out any avilable but unreleaed plaintext.
```
acioc, line 122: np: `avilable` -> `available`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+The customer MUST understand that all released plaintext is unverified until the operation
+completes and succeeds.
+The customer MUST discard any released plaintext if the operation fails, and MUST roll back
+any processing done with that released plaintext.
```
acioc, line 126: Is there a way we can enforce these statements?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+We should consider improvements to the format to further reduce or eliminate the need to release
+unverified plaintext in the streaming use case, however that is outside the scope of this document.
+
+### How and when does the header information become available?
+
+Any information from the header MUST NOT be released before the message header is authenticated.
+
+In the one-shot case, the operation MUST verify the whole message before outputting the header information.
+
+In the streaming case, the operation SHOULD release header informatoin before the message is verified,
```
acioc, line 138: np: `informatoin` -> `information`

client-apis/decrypt.md
```diff
@@ -69,65 +98,255 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+
+If deserializing [framed data](../data-format/message-body.md#framed-data),
+this operation MUST use the first 4 bytes of a frame to determine if the frame
+MUST be deserialized as a [final frame](../data-format/message-body.md#final-frame)
+or [regular frame](../fata-format/message-body/md#regular-frame).
+If the first 4 bytes have a value of 0xFFFF,
+then this MUST be deserialized as the [sequence number end](../data-format/message-header.md#sequence-number-end)
+and the following bytes according to the [final frame spec](../data-format/message-body.md#final-frame).
+Otherwise, this MUST be deserialized as the [sequence number](../data-format/message-header.md#sequence-number)
+and the following bytes according to the [regular frame spec](../data-format/message0body.md#regular-frame).
```
acioc, line 261: np: `message0body.md` -> `message-body.md`

client-apis/decrypt.md
```diff
@@ -69,65 +98,255 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  padded to the [IV length](../data-format/message-header.md#iv-length).
+- The cipherkey is the derived data key
+- The ciphertext is the [encrypted content](../data-format/message-body.md#encrypted-content).
+
+If this decryption fails, this operation MUST immediately halt and fail.
+
+If the input encrypted message is being streamed to this operation:
+
+- This operation SHOULD release the plaintext as soon as tag verification succeeds.
+  However, if this operation is using an algorithm suite with a signature algorithm
```
acioc, line 307: np: `signature algorithm` -> `signature algorithm,`

client-apis/decrypt.md
```diff
@@ -69,65 +98,255 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  all released plaintext MUST be not considered verified until
```
acioc, line 308: np: ` MUST be not considered` -> ` MUST not be considered`

client-apis/encrypt.md
```diff
@@ -83,26 +100,67 @@ A bound on the length of the [plaintext](#plaintext) to encrypt.
...
+If an implementation requires holding the entire input plaintext in memory in order to perform this operation,
+that implementation SHOULD NOT provide an API that allows this output to be streamed.
+
+## Operation
+
+The Encrypt operation is divided into several distinct parts:
+
+- [Get the encryption materials](#get-the-encryption-materials)
+- [Construct the header](#construct-the-header)
+- [Construct the body](#construc-the-body)
```
acioc, line 130: np: `construc` -> `construct`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
 Note that if the key derivation algorithm is the [identity KDF](../framework/algorithm-suites.md#identity-kdf),
 then the derived data key is the same as the plaintext data key.
 
-### Message
+The frame length used in the procedures described below is the input [frame length](#frame-size),
```
acioc, line 174: Should this link be to `../data-format/message-header.md#frame-length`?

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+if supplied.
+If a frame length is not specified on input, this operation MUST choose a reasonable value to
+default to.
```
acioc, line 177: If possible we should define what a "reasonable value" is (or at least how to determine one).

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+
+## Construct the body
+
+The encrypted message outputted by this operation MUST have a message body equal
+to the message body calculated in this step.
+
+While there MAY still be plaintext left to encrypt,
+this operation MUST either wait for more plaintext to become available,
+wait for an end to the plaintext to be indicated,
+or to perform encryption on the available plaintext to construct frames
```
acioc, line 239: Should `or to perform encryption` be `or perform encryption`? I don't think it's a `or wait to perform encryption`, but I could be wrong.

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+which make up the message body.
+
+Frames MUST be constructed sequentially such that the concatenation of their decryption
+in [sequence number](#.../data-format/message-body/md#sequence-number)
```
acioc, line 243: np: `message-body/md` -> `message-body.md`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+order is equal to the input plaintext which has been encrypted so far.
+
+If more input plaintext MAY become available
+and there are not enough input plaintext bytes available to create a new
+[regular frame](#../data-format/message-body.md#regular-frame),
+then this operation MUST wait until either enough plaintext bytes become available
+or the caller indicates an end to the plaintext.
```
acioc, line 250: How do we indicate an end to the plaintext? I believe in an earlier doc you mentioned `EOF`. Should we specify that here?

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+
+If more input plaintext MAY become available
+and there is enough plaintext available to construct a new regular frame,
+then this operation MUST [construct a regular frame](#construct-a-frame)
+with the avilable plaintext.
```
acioc, line 255: np: `avilable` -> `available`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+
+If more input plaintext MUST NOT become available
+and there is exactly enough plaintext bytes available to create one regular frame,
+then this operation MUST [construct either a final frame or regular frame](#construct-a-frame)
+with the remaining plaintext.
+If they construct a regular frame, they MUST also construct a empty final frame.
```
acioc, line 261: np: `a empty` -> `an empty`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+    Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
+  - The [sequence number](../data-format/message-body-aad.md#sequence-number) is the sequence
+    number of the frame being encrypted.
+    If this is the first frame sequentially, this value MUST be 1.
+    Otherwise, this value MUST be 1 greater than the value of the sequence number
+    of the previous frame.
+  - The [content length](../data-format/message-body-aad.md#content-length) MUST have a value
+    equal to the length of the plaintext being encrypted.
+    - For a regular frame the length of this plaintext MUST equal the frame length.
+    - For a final frame this MUST be length of the remaining plaintext bytes
```
acioc, line 291: np: `this MUST be length` `this MUST be the length`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+- The IV is the [sequence number](../data-format/message-body-aad.md#sequence-number)
+  used in the message body AAD above,
+  padded to the [IV length](../data-format/message-header.md#iv-length).
+- The cipherkey is the derived data key
+- The plaintext is the next subsequence of available plaintext bytes that have not yet been encrypted.
+  - For a regular frame the length of this plaintext subsequence MUST equal the frame length.
+  - For a final frame this MUST be the remaining plaintext bytes which have not yet been encrypted,
+    whose length MUST be equal to or less than the frame length.
+
+this operation MUST serialize a regular frame or final frame with the following specifics:
```
acioc, line 303: np: `this` -> `This`

client-apis/encrypt.md
```diff
@@ -113,68 +171,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
 
-If the [algorithm suite](../framework/algorithm-suites.md) contains a [signature algorithm](../framework/algorithm-suites.md#signature-algorithm),
-the output [message](../data-format/message.md) MUST contain a [message footer](../data-format/message-footer.md).
-The footer is serialized with the following specifics:
+- the signature key is the [signing key](../framework/structures.md#signing-key) in the [encryption materials](../framework/structures.md#encryption-materials)
+- the input to sign is the concatenation of the serialization of the [message header](../data-format/message-header.md) and [message body](../data-format/message-body.md)
+
+Note that the message header and message body MAY have already been inputted during previous steps.
+
+This operation MUST then serialize serialized a message footer with the following specifics:
```
acioc, line 335: np: `serialize serialized` -> `serialize`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
lavaleri: There isn't. This is a contract that the customer must understand for stream-decrypting messages with a message signature.

client-apis/encrypt.md
lavaleri: Should be #frame-length, referring to the input defined earlier in the spec.

client-apis/encrypt.md
lavaleri: This varies based on how streaming is done in the specific language. I updated the EOF to mirror this wording, but if we think it would be more clear to to refer to it as EOF (and maybe add it as a definition), I can do that instead.

acioc: Up to you. It might make sense to give an example of 1 or 2 languages but leave the spec as language-agnostic as possible.

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -86,7 +86,7 @@ Additionally, we use similar conditions to specify when the output of the operat
 This change SHOULD NOT introduce any language which is implementation specific for
 a specific streaming interface.
 Any streaming interface which is capable of moving bytes over time from one place to another,
-and is able to indicate EOF MUST be able to implement this specification as written.
+and is able to indicate and end to the bytes MUST be able to implement this specification as written.
```
acioc, line 89: np: `and end` -> `an end`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+
+This change rewrites these specifications to
+better specify how these operations are allowed to process data in parts.
+Specifically, we attempt to be completely language agnostic and describe streaming as a way for
+bytes to move sequentially, over time.
+
+With this framing, the things the encrypt and decrypt operations care about are:
+
+- Is the caller allowed to send any more data?
+- What data is currently available to me?
```
seebees, line 75: Why not "What data can I release?"

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+
+### How should verification and signing be performed?
+
+In order to not require holding the entire message in memory to either create
+or verify a signature, this change specifies that streaming use cases input the message
+to the signature algorithm in parts, as it is being parsed or serialized.
+
+As such, this change states that once the header or a frame is serialized or parsed,
+it MUST be inputted to the signature algorithm (if included) such that it
+isn't required to remain in memory for the duration of the operation.
```
seebees, line 99: I like having this, but should we add that if this is not possible then the implementation should not offer a streaming interface?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+signature (if included) is verified.
+
+For the streaming case, this gets more complicated.
+Streaming is a way to operate on large messages without requiring the entire message to be
+loaded into memory.
+This allows the processing of arbitrarily long messages with finite memory.
+However, when processing messages with a signature,
+the operation obtains plaintext in parts as it processes the message,
+but has no way of knowing whether the message is verified until the end of the operation.
+So what should we do with the unverified plaintext decrypted thus far?
```
seebees, line 115: What does this mean?  Is it unverified because the signature? Or because the auth tag is not verified?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+could be very expensive, when they could instead immediately see possibly unexpected data within
+the header information.
+
+This change specifies the encryption context, algorithm suite ID, and parsed header as an output,
+and makes the above security considerations clear.
+
+## Reference-level Explanation
+
+The only changes needed are to the specification,
+described by the section above.
```
robin-aws, line 154: I have been interpreting "Guide-level Explanation" as "for users of the ESDK" and this section as "for implementers of the ESDK".

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+for when the operation MAY or MUST perform some processing.
+These conditions include how many input bytes are available,
+whether more bytes MAY be available,
+at what step in the operation you are in,
+and the result of various calculations.
+
+Additionally, we use similar conditions to specify when the output of the operation MAY be released.
+
+This change SHOULD NOT introduce any language which is implementation specific for
+a specific streaming interface.
```
robin-aws, line 87: Couldn't make sense of this sentence. Do you mean "SHOULD NOT introduce any language-specific implementation..."?

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+If the input encrypted message is being streamed to this operation:
+
+- Output MUST NOT be released until otherwise indicated.
+- If no more input plaintext bytes MAY become available and this operation
+  is unable to complete the above steps with the available plaintext,
+  this operation MUST halt and indicate a failure to the caller.
+- If this operation successfully completes the above steps
+  and has processed all available input bytes,
+  this operation MAY immediately complete and succeed,
+  or it MAY wait until an end of the input bytes is indicated
```
seebees, line 166: Why do we require EOF if this is a may?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -0,0 +1,154 @@
...
+
+The following sections describe how this change is updating the specification
+to answer important questions.
+
+### How should the operation process data in parts?
+
+This change rewrites these specifications to
+better specify how these operations are allowed to process data in parts.
+Specifically, we attempt to be completely language agnostic and describe streaming as a way for
+bytes to move sequentially, over time.
```
robin-aws, line 70:
```suggestion
bytes to be read and written sequentially, over time.
```

"to move" sounds too much like the bytes are moving "on their own", implying concurrency.

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+The data key used as input for all decryption described below is a data key derived from the plaintext data key
+included in the [decryption materials](../framework/structures.md#decryption-materials).
+The algorithm used to derive a data key from the plaintext data key MUST be
+the [key derivation algorithm](../framework/algorithm-suites.md#key-derivation-algorithm) included in the
+[algorithm suite](../framework/algorithm-suites.md) defined above.
+This document refers to the output of the key derivation algorithm as the derived data key.
+Note that if the key derivation algorithm is the [identity KDF](../framework/algorithm-suites.md#identity-kdf),
+then the derived data key is the same as the plaintext data key.
+
+### Verify the header
```
seebees, line 219: This and the Body verification should be moved to the Algorithm Suites, but that can be done in a later feature request.

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
-  a [final frame](../data-format/message-body.md#final-frame) or is [non framed](../data-format/message-body.md#non-framed-data).
-  Refer to [Message Body AAD](../data-format/message-body-aad.md) specification for more information.
-- Sequence Number: This value is the [sequence number](../data-format/message-body.md#sequence-number) of the frame being decrypted, if the message contains framed data.
-  If the message contains non framed data, then this value is 1.
-- Content Length: TODO
+A collection of deserialized fields of the [encrypted message's](#encrypted-message) header.
 
-If the algorithm suite has a signature algorithm, decrypt MUST verify the message footer using the specified signature algorithm,
-by using the verification key obtained from the decryption materials.
+## Operation
```
robin-aws, line 141: Behaviour instead?

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+If any of these steps fails, this operation MUST fail and indicate a failure to the caller.
 
-## Apendix
+If the input encrypted message is not being streamed to this operation,
+all output MUST NOT be released until after these steps complete successfully.
+
+If the input encrypted message is being streamed to this operation:
+
+- Output MUST NOT be released until otherwise indicated.
+- If no more input plaintext bytes MAY become available and this operation
```
robin-aws, line 160: Would read better as "if all input bytes have been provided"

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  is unable to complete the above steps with the available plaintext,
+  this operation MUST halt and indicate a failure to the caller.
+- If this operation successfully completes the above steps
+  and has processed all available input bytes,
+  this operation MAY immediately complete and succeed,
+  or it MAY wait until an end of the input bytes is indicated
+  to complete and succeed.
```
robin-aws, line 167: I don't get this one. How can it succeed if it doesn't know if more input is coming and hence hitting the case immediately below?

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
-## Apendix
+If the input encrypted message is not being streamed to this operation,
+all output MUST NOT be released until after these steps complete successfully.
+
+If the input encrypted message is being streamed to this operation:
+
+- Output MUST NOT be released until otherwise indicated.
+- If no more input plaintext bytes MAY become available and this operation
+  is unable to complete the above steps with the available plaintext,
+  this operation MUST halt and indicate a failure to the caller.
```
mattsb42-aws, line 162: you mean ciphertext? this being decrypt

client-apis/decrypt.md
```diff
@@ -69,65 +98,259 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+  However, if this operation is using an algorithm suite with a signature algorithm
+  all released output MUST be not considered verified until
+  this operation successfully completes.
+  See [security considerations](#security-considerations) below.
+- This operation SHOULD input the serialized header to the signature algorithm as soon as it is deserialized,
+  such that the serialized frame isn't required to remain in memory to complete
+  the [signature calculation](#signature-calculation).
+
+### Decrypt the message body
+
```
mattsb42-aws, line 248: Implied by some other bits of this, but the plaintext of a frame MUST NOT be released until it has been authenticated.

client-apis/decrypt.md
```diff
@@ -5,7 +5,17 @@
 
 ## Version
 
-0.1.0-preview
+0.2.0
+
+## Changlog
```
alex-chew, line 10: s/changlog/changelog/

client-apis/decrypt.md
lavaleri: Deciding to cut this wording out for now, as this is more of a question of how implementations choose to interpret streaming input. Will determine offline if there are any additional constraints we should add in the Input section above. But for now this should be covered by "there is some mechanism to determine end of input", and how each implementation defines that mechanism may differ. For Python, the mechanism is the operation itself determining that it has received a valid encrypted message, and opting to not read any more bytes.

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -59,20 +78,42 @@ There SHOULD be no operational implications for specifying the streaming use cas
...
+However, in the case of streaming messages with signatures,
+any plaintext released by the Decrypt operation MUST NOT be
+considered signed data until the operation completes and succeeds.
+
+This means that customers that process such released plaintext MUST NOT consider any processing successful
+until this operation completes successfully.
+Additionally, if this operation fails, customers MUST discard the released plaintext and encryption context
+and MUST rollback any processing done due to the released plaintext or encryption context.
+
+This change adds these security considerations are stated in the Decrypt specification.
```
acioc, line 98: np: `as stated`?

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -111,32 +166,33 @@ loaded into memory.
...
 Because streaming is being used, we MUST NOT buffer the entire plaintext,
 as that would defeat the point of streaming.
 Thus, the plaintext needs to go somewhere other than memory.
 Because this is a client-side library, we MUST give it to the customer.
 
 In the case of any decryption failure,
 the operation MUST zero out any available but unreleaed plaintext.
-The customer MUST understand that all released plaintext is unverified until the operation
-completes and succeeds.
+The customer MUST understand that all released plaintext is cannot be considered signed data
```
acioc, line 178: np: `all released plaintext cannot...`

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -111,32 +166,33 @@ loaded into memory.
...
+until the operation completes and succeeds.
 The customer MUST discard any released plaintext if the operation fails, and MUST roll back
 any processing done with that released plaintext.
 This change states these security considerations.
 
 We should consider improvements to the format to further reduce or eliminate the need to release
-unverified plaintext in the streaming use case, however that is outside the scope of this document.
+of such plaintext in the streaming use case, however that is outside the scope of this document.
```
acioc, line 185: np: maybe `such` instead of `of such`

client-apis/decrypt.md
```diff
@@ -237,7 +259,7 @@ If the input encrypted message is being streamed to this operation:
   and [other header information](#parsed-header)
   as soon as tag verification succeeds.
   However, if this operation is using an algorithm suite with a signature algorithm
-  all released output MUST be not considered verified until
+  all released output MUST be not considered signed data until
```
acioc, line 262: np: `MUST NOT be`

client-apis/encrypt.md
```diff
@@ -60,6 +64,19 @@ and MUST be ignored.
...
+This MUST be a sequence of bytes.
+
+This input MAY be streamed to this operation.
+This means that bytes are made available to this operation sequentially and over time,
+and that an end to the input is eventually indicated.
+This means that:
+
+- There MUST be a mechanism for input bytes to become available to be processed.
+- There MUST be a mechanism to indicate that there are no more input bytes,
+  or whether more bytes MAY be made available in the future.
```
robin-aws, line 76: I think it's worth lifting this definition into its own client-apis/streaming.md feature, or perhaps adding to structures.md. You can focus there on terminology definitions (things like the "EOF" that is a separate signal and NOT an EOF marker, the idea of "releasing" data, etc), describe more explicitly what it means for an operation to take a stream as input vs. produce a stream as output, and so on. It will also help shorten and focus the content here.

client-apis/decrypt.md
robin-aws: Big +1 from me for factoring as much as possible into Algorithm Suites to shorten these features. If possible to do in this PR I'd find it easier to review and sign-off on the whole package.

client-apis/encrypt.md
```diff
@@ -113,68 +177,185 @@ This document refers to the output of the key derivation algorithm as the derive
...
+this operation MUST either wait for more plaintext to become available,
+wait for an end to the plaintext to be indicated,
+or perform encryption on the available plaintext to construct frames
+which make up the message body.
+
+Frames MUST be constructed sequentially such that the concatenation of their decryption
+in [sequence number](#.../data-format/message-body.md#sequence-number)
+order is equal to the input plaintext which has been encrypted so far.
+
+If more input plaintext MAY become available
```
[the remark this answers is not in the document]
lavaleri: I've updated this section to reframe it as "before EOF, process as many available bytes into regular frames as you can", "after EOF, follow these steps to process the remaining bytes".

client-apis/encrypt.md
lavaleri: Moved these concerns to client-api/streaming.md

client-apis/decrypt.md
lavaleri: As agreed offline, we will do this in a separate PR scoped to spec 1.1. Created issue here: https://github.com/awslabs/aws-encryption-sdk-specification/issues/162

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
+
+### Available Bytes
+
+In the scope of an operation, bytes are considered available if
+the operation has not yet processed those bytes
+and those bytes are intended to be processed,
+this intention being expressed through the specific streaming interface.
```
acioc, line 40:
```suggestion
and those bytes are intended to be processed.
This intention is expressed through the specific streaming interface.
```

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+and that operation must read from some source in order to produce bytes,
+bytes read from the source are considered available and MUST be processed.
+Here it is the operation expressing intent to process a set of input bytes.
+Even if more bytes MAY be read from the source,
+bytes are not considered available to an operation
+until they have been consumed from the source.
+
+In a framework where a customer is sending input bytes to an operation
+and that operation must write those bytes to some sink,
+the bytes recieved from the customer are considered available and MUST be processed.
```
acioc, line 52:
```suggestion
the bytes received from the customer are considered available and MUST be processed.
```

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+For example, in a framework where a customer is requesting output bytes from an operation
+and that operation must read from some source in order to produce bytes,
+bytes read from the source are considered available and MUST be processed.
+Here it is the operation expressing intent to process a set of input bytes.
+Even if more bytes MAY be read from the source,
+bytes are not considered available to an operation
+until they have been consumed from the source.
+
+In a framework where a customer is sending input bytes to an operation
+and that operation must write those bytes to some sink,
```
acioc, line 51: This is probably fine, but it *might* be worth either explaining a sink or using another word for it? It is fairly typical streaming terminology though, so up to you.

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+  or whether more bytes MAY be made available in the future.
+
+These mechanisms are used to allow the operation to process input bytes in parts, over time.
+
+The bytes that represent the entire input to the operation are the bytes that were made available
+up until an end was indicated or the operation is able to determine
+that the bytes made available represent the full input.
+
+The [encrypt](encrypt.md) operation is unable to determine whether it has processed the entire
+input on it's own, and thus MUST NOT complete until an end to the input is indicated.
```
acioc, line 81:
```suggestion
input on its own, and thus MUST NOT complete until an end to the input is indicated.
```

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+If an operation is able to determine that is has recieved the full input without an end being
```
acioc, line 83:
```suggestion
If an operation is able to determine that is has received the full input without an end being
```

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+In order to support streaming, the operation MUST return some output within a streaming framework.
+
+This means that:
+
+- There MUST be a mechanism for output bytes to be released
+- There MUST be a mechanism to indicate that the entire output has been released,
+  or whether more bytes MAY be released in the future.
+
+These mechanisms are used to allow the operation output processed bytes in parts, over time.
```
acioc, line 96:
```suggestion
These mechanisms are used to allow the operation output to process bytes in parts, over time.
```

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Behavior
+
+By using the mechanisms for [inputs](#inputs) and [outputs](#outputs),
+some actor expresses intent through a streaming interface
+for bytes to be made available to the operation
+and for bytes to be released by the operation.
+
+The behavior of the operation specifies how the operation processes available bytes,
+and specifies when processed bytes can be released.
```
acioc, line 111:
```suggestion
and specifies when processed bytes MAY be released.
```
I personally prefer MAY instead of can to match our other terminology.

client-apis/streaming.md
lavaleri: Unless we can agree on a better word, I think "sink" here is sufficiently clear with it's intention here, especially when paired with "source" above.

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+that implementation SHOULD NOT provide an API that allows the caller to stream the operation.
+
+## Definitions
+
+### Conventions used in this document
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
+
+### Available Bytes
```
robin-aws, line 35: This term might be a little overloaded - your description is right in that just because bytes COULD be read doesn't mean they are INTENDED to be. I ended up using "consumable" in my comments to mean the former instead.

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+In order to support streaming, the operation MUST return some output within a streaming framework.
+
+This means that:
+
+- There MUST be a mechanism for output bytes to be released
+- There MUST be a mechanism to indicate that the entire output has been released,
+  or whether more bytes MAY be released in the future.
+
+These mechanisms are used to allow the operation output to process bytes in parts, over time.
```
robin-aws, line 96:
```suggestion
These mechanisms are used to allow the operation output to produce bytes in parts, over time.
```
?

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+The bytes that represent the entire output to the operation are the bytes that were released
+up until an end was indicated.
+
+Operations MUST NOT indicate completion or success until an end to the output has been indicated.
```
robin-aws, line 101: Love this statement.

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+The [encrypt](encrypt.md) operation is unable to determine whether it has processed the entire
+input on its own, and thus MUST NOT complete until an end to the input is indicated.
+
+If an operation is able to determine that is has received the full input without an end being
+indicated, the operation MUST fail if more bytes become available.
+
+## Outputs
+
+In order to support streaming, the operation MUST return some output within a streaming framework.
```
robin-aws, line 88:
```suggestion
In order to support streaming, the operation MUST produce some output within a streaming framework.
```
I definitely wouldn't use "return" here since that implies the synchronous result of a function/procedure.

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+In a framework where a customer is sending input bytes to an operation
+and that operation must write those bytes to some sink,
+the bytes received from the customer are considered available and MUST be processed.
+Here the customer is expressing intent to process their supplied bytes.
+
+### Release
+
+An operation releases processed bytes when those bytes are outputted by the operation.
+
+If bytes are processed by an operation, that does not imply that the operation is allowed to
```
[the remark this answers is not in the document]
lavaleri: I mean process here. i.e. you deserialized the bytes of the header, you still can't release them until the header auth is checked. "produce" seems to imply "output", which doesn't make sense here.

client-apis/encrypt.md
```diff
@@ -256,42 +242,28 @@ the [signature calculation](#signature-calculation).
...
+When it is indicated that no more plaintext will be available,
+this operation MUST perform the following until all available plaintext is processed:
+
+- If there are enough input plaintext bytes available to create a new regular frame,
+  then this operation MUST [construct a regular frame](#construct-a-frame).
+  using the available plaintext.
+- If there are exactly enough plaintext bytes available to create one regular frame,
+  then this operation MUST [construct either a final frame or regular frame](#construct-a-frame)
+  with the remaining plaintext.
+  If they construct a regular frame, they MUST also construct an empty final frame.
```
robin-aws, line 263: There's actually a bug here with the edge case of an empty plaintext: technically the while loop immediately terminates and you don't require an empty final frame.

It may be better to move this line to after the loop and state "If a final frame has not yet been constructed, they MUST construct an empty final frame"

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+### Available Bytes
+
+In the scope of an operation, bytes are considered available if
+the operation has not yet processed those bytes
+and those bytes are intended to be processed.
+This intention is expressed through the specific streaming interface.
+
+For example, in a framework where a customer is requesting output bytes from an operation
+and that operation must read from some source in order to produce bytes,
+bytes read from the source are considered available and MUST be processed.
```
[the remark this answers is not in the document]
lavaleri: That's a good way to reframe this, but I was having a hard time framing this such that it still made sense when you thought about how the bytes you want to read from source might not be ready to be read yet. I've reworked this part a bit. Thoughts?

client-apis/streaming.md
acioc: That's totally fine. Let's keep "sink".

client-apis/streaming.md
```diff
@@ -0,0 +1,111 @@
...
+
+- There MUST be a mechanism for input bytes to become available to be processed.
+- There MUST be a mechanism to indicate that there are no more input bytes,
+  or whether more bytes MAY be made available in the future.
+
+These mechanisms are used to allow the operation to process input bytes in parts, over time.
+
+The bytes that represent the entire input to the operation are the bytes that were made available
+up until an end was indicated or the operation is able to determine
+that the bytes made available represent the full input.
```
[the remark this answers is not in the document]
lavaleri: Good call. Updated and cut out the explanation around operations trying to determine the end by themselves. It's implied in the users intent and it's the streaming framework responsibility to express that intent to the operation.

client-apis/streaming.md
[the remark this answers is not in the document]
robin-aws: True. I was more thinking "if bytes are ready to be released by the operation". Thinking more generically, a serialization operation doesn't process bytes as input, but does produce them in a stream. Since our only instances of streaming to date have streaming input AND output this isn't a major point to worry about.

client-apis/encrypt.md
```diff
@@ -256,42 +242,28 @@ the [signature calculation](#signature-calculation).
...
+this operation MUST process as much of the available bytes as possible
+by [constructing regular frames](#construct-a-frame).
+
+When it is indicated that no more plaintext will be available,
+this operation MUST perform the following until all available plaintext is processed:
+
+- If there are enough input plaintext bytes available to create a new regular frame,
+  then this operation MUST [construct a regular frame](#construct-a-frame).
+  using the available plaintext.
+- If there are exactly enough plaintext bytes available to create one regular frame,
```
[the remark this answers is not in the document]
lavaleri: ah, this should be a strict `>`, `<`, vs `==`. which means I should update the first to specify exactly `>` (and reorder so the == is first).

client-apis/streaming.md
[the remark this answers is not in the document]
robin-aws: Actually, I take that back. :) I was allowing for the fact that you can refer to "If input bytes are processed, which means the operation has bytes to produce as output" which is not precisely correct in all cases. Personally I interpret "bytes are processed" as meaning the bytes are input and NOT output.

client-apis/streaming.md
```diff
@@ -0,0 +1,112 @@
...
+when they SHOULD be released,
+and when they MUST NOT be released.
+
+## Inputs
+
+In order to support streaming, the operation MUST accept some input within a streaming framework.
+
+This means that:
+
+- There MUST be a mechanism for input bytes to become consumable.
```
robin-aws, line 79: Perhaps imply a little more strongly that this is referring to a specific sequence of bytes at a time, and happens repeatedly? Otherwise it reads a bit as "a mechanism for bytes to flow in some way over time"

client-apis/streaming.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Outputs
+
+In order to support streaming, the operation MUST produce some output within a streaming framework.
+
+This means that:
+
+- There MUST be a mechanism for output bytes to be released
+- There MUST be a mechanism to indicate that the entire output has been released,
+  or whether more bytes MAY be released in the future.
```
robin-aws, line 95: Same comment as the one you accepted on Inputs here.

client-apis/streaming.md
robin-aws: Should the header be "Consumable Bytes" too then?

client-apis/streaming.md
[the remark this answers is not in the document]
lavaleri: When I say "processed bytes" I mean, the result of processing some input bytes, not the input bytes that were processed. I can see how that would be confusing here.
I have updated to mention the result of processing.

client-apis/encrypt.md
```diff
@@ -247,23 +247,28 @@ and this operation determines at any time that the plaintext being encrypted
 has a length greater than this value,
 this operation MUST immediately fail.
 
-Before it is indicated that no more plaintext will be available,
-this operation MUST process as much of the available bytes as possible
+Before and end to the input is indicated,
```
acioc, line 250:
```suggestion
Before the end of the input is indicated,
```

client-apis/encrypt.md
```diff
@@ -247,23 +247,28 @@ and this operation determines at any time that the plaintext being encrypted
...
+this operation MUST process as much of the consumable bytes as possible
 by [constructing regular frames](#construct-a-frame).
 
-When it is indicated that no more plaintext will be available,
-this operation MUST perform the following until all available plaintext is processed:
+When end to the input is indicated,
```
acioc, line 254:
```suggestion
When the end of the input is indicated,
```

client-apis/streaming.md
```diff
@@ -32,33 +34,40 @@ that implementation SHOULD NOT provide an API that allows the caller to stream t
...
+and that operation must read from some source in order to produce bytes,
+this is slightly more complicated.
+Bytes are considered consumable if:
+
+- Those bytes have not yet been processed.
+- Those bytes are able to be read by the operation from the source.
+- Those bytes are required to be processed in order for the operation
+  to release the output requested by the customer.
+  Here the customer expresses intent for the operation to process
+  whatever the operation needs to consume to produce its complete output
```
acioc, line 61: Wanted to double check the indent here.

client-apis/streaming.md
```diff
@@ -32,33 +34,40 @@ that implementation SHOULD NOT provide an API that allows the caller to stream t
...
 
-For example, in a framework where a customer is requesting output bytes from an operation
-and that operation must read from some source in order to produce bytes,
-bytes read from the source are considered available and MUST be processed.
-Here it is the operation expressing intent to process a set of input bytes.
-Even if more bytes MAY be read from the source,
-bytes are not considered available to an operation
-until they have been consumed from the source.
+- The operation has not yet processed those bytes,
+- The operation has access to those bytes,
```
acioc, line 42:
```suggestion
- The operation has not yet processed those bytes.
- The operation has access to those bytes.
```

client-apis/streaming.md
```diff
@@ -32,33 +34,40 @@ that implementation SHOULD NOT provide an API that allows the caller to stream t
...
+- Those bytes are intended to be processed.
+  This intention is expressed through the specific streaming interface.
```
acioc, line 44: Wanted to double check this indent

client-apis/streaming.md
```diff
@@ -67,33 +76,24 @@ In order to support streaming, the operation MUST accept some input within a str
...
+to be processed.
 
 ## Outputs
 
-In order to support streaming, the operation MUST return some output within a streaming framework.
+In order to support streaming, the operation MUST produce some output within a streaming framework.
 
 This means that:
 
 - There MUST be a mechanism for output bytes to be released
```
acioc, line 93:
```suggestion
- There MUST be a mechanism for output bytes to be released.
```

client-apis/encrypt.md
```diff
@@ -60,6 +64,12 @@ and MUST be ignored.
 ### Plaintext
 
 The plaintext to encrypt.
+This MUST be a sequence of bytes.
```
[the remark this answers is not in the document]
robin-aws: See my comment on this here: https://github.com/awslabs/aws-encryption-sdk-specification/pull/154#discussion_r455248406

Multiple places in the specification should be able to say X "is a sequence of bytes" without getting into these details over and over again. IF we want to get into those details they should be in a top level concept feature similar to streaming.md.

client-apis/decrypt.md
```diff
@@ -69,65 +105,260 @@ This default CMM MUST obtain the decryption materials required for decryption.
...
+    of the previous frame.
+  - The [content length](../data-format/message-body-aad.md#content-length) MUST have a value
+    equal to the length of the plaintext that was encrypted.
+    This can be determined by using the [frame length](../data-format/message-header.md#frame-length)
+    deserialized from the message header if this is a regular frame,
+    or the [encrypted content length](../data-format/message-body.md#encrypted-content-length)
+    otherwise.
+- The IV is the [sequence number](../data-format/message-body-aad.md#sequence-number)
+  used in the message body AAD above,
+  padded to the [IV length](../data-format/message-header.md#iv-length).
```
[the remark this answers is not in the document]
robin-aws: You still want the "to " though?

client-apis/encrypt.md
```diff
@@ -130,72 +180,180 @@ This document refers to the output of the key derivation algorithm as the derive
...
+    of the previous frame.
+  - The [content length](../data-format/message-body-aad.md#content-length) MUST have a value
+    equal to the length of the plaintext being encrypted.
+    - For a regular frame the length of this plaintext MUST equal the frame length.
+    - For a final frame this MUST be the length of the remaining plaintext bytes
+      which have not yet been encrypted,
+      whose length MUST be equal to or less than the frame length.
+- The IV is the [sequence number](../data-format/message-body-aad.md#sequence-number)
+  used in the message body AAD above,
+  padded to the [IV length](../data-format/message-header.md#iv-length).
```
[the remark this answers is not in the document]
robin-aws: You still want the "to " though?

client-apis/encrypt.md
```diff
@@ -130,72 +180,180 @@ This document refers to the output of the key derivation algorithm as the derive
...
+
+- [Sequence Number](../data-format/message-body.md#sequence-number): MUST be the sequence number of this frame,
+  as determined above.
+- [IV](../data-format/message-body.md#iv): MUST be the IV used when calculating the encrypted content above
+- [Encrypted Content](../data-format/message-body.md#encrypted-content): MUST be the encrypted content calculated above.
+- [Authentication Tag](../data-format/message-body.md#authentication-tag): MUST be the authentication tag
+  outputted when calculating the encrypted content above.
+
+Once the entire frame is serialized,
+the serialized frame MAY be released.
```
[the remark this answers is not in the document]
robin-aws: I think the more important statement here is the implicit opposite: "The serialized frame MUST NOT be released before the entire frame is serialized". Perhaps it would be better to state that instead.

Given that, I'm not sure we need to weigh in on the MAY vs. SHOULD part on when it IS released.

client-apis/encrypt.md
```diff
@@ -130,72 +180,180 @@ This document refers to the output of the key derivation algorithm as the derive
...
-  - the input to sign is the concatenation of the serialization of the [message header](../data-format/message-header.md) and [message body](../data-format/message-body.md)
+This operation MUST then serialize a message footer with the following specifics:
 
-Any data that is not specified within the [message format](../data-format/message.md) MUST NOT be added to the output message.
+- [Signature Length](../data-format/message-footer.md#signature-length): MUST be the length of the
+  output of the calculation above.
+- [Signature](../data-format/message-footer.md#signature): MUST be the output of the calculation above.
+
+Once the entire message is signed and the message footer is serialized,
+the serialized message footer MAY be released.
```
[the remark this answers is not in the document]
robin-aws: Same as above, although in this case it feels like this statement and the next one could just be combined into one, since there's only one correct moment in time to release this.

changes/2020-07-06_clarify-streaming-encrypt-decrypt/change.md
```diff
@@ -166,8 +214,6 @@ as that would defeat the point of streaming.
 Thus, the plaintext needs to go somewhere other than memory.
 Because this is a client-side library, we MUST give it to the customer.
 
-In the case of any decryption failure,
-the operation MUST zero out any available but unreleaed plaintext.
```
lavaleri: Just a note that I cut this out now that I've slightly tweaked the definition for release (also realized that this was in the change doc but not the spec). I'm opting to not prescribe what to do for failure cases since what you can or should do will depend on the streaming framework.

client-apis/decrypt.md
```diff
@@ -98,6 +98,26 @@ This input MAY be [streamed](streaming.md) to this operation.
...
 that implementation SHOULD NOT provide an API that allows the caller to stream the encrypted message.
 
+#### Encrypted Message Format
+
+The message format is a binary format, but it is a common mistake for users to attempt decryption on the Base64 encoding of this data instead.
+Because the first two bytes of the message format have a very limited set of possible values
+(currently they are in fact fixed),
+the first two bytes of the Base64 encoding of a valid message are also simple to recognize.
+
+To make diagnosing this mistake easier, implementations SHOULD detect the first two bytes of the Base64 encoding of any supported message [versions](../data-format/message-header.md#version-1)
```
acioc:
```suggestion
To make diagnosing this mistake easier,
implementations SHOULD detect the first two bytes of the Base64 encoding
of any supported message [versions](../data-format/message-header.md#version-1)
```
np: to split this long line

client-apis/decrypt.md
```diff
@@ -325,14 +346,20 @@ This operation MUST NOT release any unauthenticated plaintext.
...
 
-- This operation SHOULD release the plaintext as soon as tag verification succeeds.
+- This operation SHOULD release plaintext decrypted from unframed data or regular frames
+  as soon as tag verification succeeds.
   However, if this operation is using an algorithm suite with a signature algorithm,
   all released plaintext MUST NOT be considered signed data until
   this operation successfully completes.
   See [security considerations](#security-considerations) below.
+- If this operation is using an algorithm suite with a signature algorithm,
+  this opeartion MUST NOT release any plaintext decrypted from a final frame until
```
acioc, line 356:
```suggestion
  this operation MUST NOT release any plaintext decrypted from a final frame until
```

client-apis/encrypt.md
```diff
@@ -223,13 +223,15 @@ With the authentication tag calculated, this operation MUST serialize the
...
+- [IV](../data-format/message-header.md#iv): MUST have the value of the IV used in the calculation above,
+  padded to the [IV length](../data-format/message-header.md#iv-length) with 0.
 - [Authentication Tag](../data-format/message-header.md#authentication-tag): MUST have the value
   of the authentication tag calculated above.
 
-Once the entire message header has been authenticated and serialized,
-the serialized message header MAY be released.
+The serialized bytes MUST NOT be released until the entire message header has been serialized
+If this operation is streaming the encrypted message and
+once the entire message header has been serialized,
```
acioc, line 233:
```suggestion
the entire message header has been serialized,
```

client-apis/encrypt.md
```diff
@@ -316,8 +318,10 @@ This operation MUST serialize a regular frame or final frame with the following
 - [Authentication Tag](../data-format/message-body.md#authentication-tag): MUST be the authentication tag
   outputted when calculating the encrypted content above.
 
-Once the entire frame is serialized,
-the serialized frame MAY be released.
+The above serialized bytes MUST NOT be released until the entire frame has been serialized.
+If this operation is streaming the encrypted message and
+once the entire frame has been serialized,
```
acioc, line 323:
```suggestion
the entire frame has been serialized,
```

client-apis/encrypt.md
```diff
@@ -95,17 +105,55 @@ greater than this value.
 
 This behavior MUST output the following if the behavior is successful:
 
-- [message](../data-format/message.md)
+- [encrypted message](#encrypted-message)
```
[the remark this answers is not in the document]
seebees: Yes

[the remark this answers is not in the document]
lavaleri: Done.

PR #152, opened 2020-07-13
fix: Replace “children keyrings” with “child keyrings”

### LINKED ISSUE #115: Multi keyring child keyrings vs children keyrings
Should the interface be changed to `child` to simplify the documentation grammar?

robin-aws:
Fixes #115

Also a couple of minor grammar/readability fixes.

Note that renaming the sub-heading breaks the existing #children-keyrings anchor. I'm making the assumption that this is allowable given all current references are within this feature document. If that's not the case I can add an explicit anchor with the old symbol.


robin-aws:
> This looks good, but what is our plan to "fix" this?
> Is it a requirement because if so this is a breaking change?

As noted in #115 the unreleased Java implementation is the only one that needs fixing, so I planned to cut a follow-up issue once this was merged. Come to think of it I can do that ahead of time though.

Issue #137, opened 2020-06-17
document spec change doc format and process


mattsb42-aws:
Before we define this as "1.0", we need to document how to make changes to the spec and what process we use to take a change from idea to proposal to acceptance.

We've been roughly following a sketch of a process that I put together before we had an opportunity to try following it, and unsurprisingly, we've been finding places where it needs to change before being finalized here.

Points we need to hit include:

- [ ] Templates for changes
- [ ] Templates for PRs
- [ ] Templates for issues
- [ ] Repo organization (where to changes go, how to they reflect on spec docs, etc)
- [ ] Process for accepting a change into the spec
- [ ] Process for accepting a proposal that needs more review before integrating into the spec

mattsb42-aws:
## Organization

The general pattern that we have settled on for repo organization is:

- Spec documents (currently in "client-apis", "data-format", and "framework") describe the current state of the specification. Each spec document MUST contain a version and a changelog that describes what changed in each version with a link to the accepted change document that introduced that change.
- Change documents describe a change to the specification. Each change document MAY affect multiple spec documents.
- Accepted change documents are found in the "changes" directory.
- Each change has a dedicated directory with a name matching the pattern `YYYY-MM-DD_NAME`, identifying the date when the change was proposed and a name that summarizes the change.

Open questions:

- [ ] Do we want to collect all spec documents under a common parent directory to more clearly differentiate them from changes/etc?
- [ ] Where do proposed changes go?

mattsb42-aws:
## Templates for changes

The template that I have been using is:


```markdown
[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"

# TITLE GOES HERE

## Affected Features

This serves as a reference of all features that this change affects.

| Feature |
| ------- |


## Affected Specifications

This serves as a reference of all specification documents that this change affects.

| Specification |
| ------------- |


## Affected Implementations

This serves as a reference for all implementations that this change affects.

| Language   | Repository                                                                            |
| ---------- | ------------------------------------------------------------------------------------- |
| Python     | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-python)              |
| Java       | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-java)       |
[...]

PR #154, opened 2020-07-13
feat: Add recommendation for detecting base64-encoded messages

### LINKED ISSUE #42: Verification of version and type on deserialization base64 encoding error
When deserializing an encrypted blob.
The version and type are fixed.
However binary data is often stored as a base64 encoded string.
This is especially true of the AWS Encryption SDK Java functions `encryptString`.

Customers will take a base64 encoded string and pass this to a `decrypt` function without the proper decoding, resulting in a failure.

Since the version and type are fixed,
and these values can be translated into their base64 equivalent,
when the blob is rejected as invalid,
the notification can suggest that this failure is an encoding issue.

This may be codified into the specification
or as a recommendation for implementations.

Changes needed:

- [ ] Specification
- [ ] Javascript:
https://github.com/awslabs/aws-encryption-sdk-javascript/pull/217
- [ ] C
https://github.com/aws/aws-encryption-sdk-c/issues/445
- [ ] CLI
https://github.com/aws/aws-encryption-sdk-cli/issues/162
- [ ] Java
https://github.com/aws/aws-encryption-sdk-java/issues/124
- [ ] Python
https://github.com/aws/aws-encryption-sdk-python/issues/192

robin-aws:
Fixes #42 

### DIFF
--- a/changes/2020-07-13_detect-base64-encoded-messages/change.md
+++ b/changes/2020-07-13_detect-base64-encoded-messages/change.md
@@ -10,7 +10,6 @@ This serves as a reference of all features that this change affects.
 | Feature                                                                                                                                     |
 | ------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Decrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/623992d999db0b309d8a8adbd664f0d72feee813/client-apis/decrypt.md) |
-|                                                                                                                                             |
 
 ## Affected Specifications
 
@@ -52,10 +51,11 @@ the first two bytes of the Base64 encoding of a valid message are also simple to
 
 ## Drawbacks
 
-This feature technically means the version value is capped to 40 or less,
-to ensure that the Base64-encoded values are not ambiguous.
+This feature technically means the version value
+cannot be incremented beyond the hex value `40`,
+since `41` would then conflict with the Base64-encoded value for `1`.
 This does not seem worth worrying about
-given how rarely we intended to change the version.
+given how rarely we intended to change the version, however.
 
 ## Security Implications
 
--- a/changes/2020-07-13_detect-base64-encoded-messages/change.md
+++ b/changes/2020-07-13_detect-base64-encoded-messages/change.md
@@ -7,9 +7,9 @@
 
 This serves as a reference of all features that this change affects.
 
-| Feature                                                                                                                                     |
-| ------------------------------------------------------------------------------------------------------------------------------------------- |
-| [Decrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/623992d999db0b309d8a8adbd664f0d72feee813/client-apis/decrypt.md) |
+| Feature                                 |
+| --------------------------------------- |
+| [Decrypt](../../client-apis/decrypt.md) |
 
 ## Affected Specifications
 
--- a/changes/2020-07-13_detect-base64-encoded-messages/change.md
+++ b/changes/2020-07-13_detect-base64-encoded-messages/change.md
@@ -10,7 +10,7 @@ This serves as a reference of all features that this change affects.
 | Feature                                                                                                                                     |
 | ------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Decrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/623992d999db0b309d8a8adbd664f0d72feee813/client-apis/decrypt.md) |
-|  |
+|                                                                                                                                             |
 
 ## Affected Specifications
 
--- a/changes/2020-07-13_detect-base64-encoded-messages/change.md
+++ b/changes/2020-07-13_detect-base64-encoded-messages/change.md
@@ -1,0 +1,88 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Detect Base64-encoded Messages
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                                                                                                                                            |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| [Decrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/623992d999db0b309d8a8adbd664f0d72feee813/client-apis/decrypt.md) |
+                                                         |
+
+## Affected Specifications
+
+This serves as a reference of all specification documents that this change affects.
+
+| Specification                                                                 |
+| ----------------------------------------------------------------------------- |
+| [Decrypt](../../client-apis/decrypt.md)                                       |
+
+## Affected Implementations
+
+This serves as a reference for all implementations that this change affects.
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
+| Java       | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-java)                |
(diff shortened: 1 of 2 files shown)


robin-aws:
Realized I still need a change doc and version change etc. for this, coming soon.

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+## Definitions
+
+### Conventions used in this document
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Summary
+
+When a decryption attempt is made on the Base64-encoding of a valid message,
```
MatthewBennington, line 43: > a decryption attempt is made on the Base64-encoding of a valid message

Does this mean a user is trying to decrypt a Base64-string, the utf-8 encoding of a Base64-string, or both/either?

robin-aws: The UTF-8 encoding of the string - decrypt accepts a message as a byte sequence after all.

FWIW, a Base64 string only uses ASCII characters so nearly all encodings will use the same byte values.

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+# Detect Base64-encoded Messages
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                                                                                                                     |
+| ------------------------------------------------------------------------------------------------------------------------------------------- |
+| [Decrypt](https://github.com/awslabs/aws-encryption-sdk-specification/blob/623992d999db0b309d8a8adbd664f0d72feee813/client-apis/decrypt.md) |
+|                                                                                                                                             |
```
acioc, line 13: np: extra row

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Affected Specifications
+
+This serves as a reference of all specification documents that this change affects.
+
+| Specification                           |
+| --------------------------------------- |
+| [Decrypt](../../client-apis/decrypt.md) |
```
acioc, line 21: Should the SHA be here?

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+## Motivation
+
+The message format is a binary format, but it is a common mistake for users to attempt decryption on the Base64 encoding of this data instead.
+Because the first two bytes of the message format have a very limited set of possible values
+(currently they are in fact fixed),
+the first two bytes of the Base64 encoding of a valid message are also simple to recognize.
+
+## Drawbacks
+
+This feature technically means the version value is capped to 40 or less,
```
acioc, line 55: 40 what?

client-apis/decrypt.md
```diff
@@ -7,6 +7,16 @@
 
 0.1.0-preview
 
+### Changelog
+
+- 0.2.0
+
+  - [Detect Base64-encoded Messages](../changes/2020-05-13_remove-keyring-trace/change.md)
```
acioc, line 14: Should this be `../changes/2020-07-13_detect-base64-encoded-messages/change.md` ?

robin-aws: Yup, copy/paste error.

changes/2020-07-13_detect-base64-encoded-messages/change.md
robin-aws: As in literally the version must be <= 40. If we ever bumped it to 41 it would be ambiguous with the Base64 encoding of the version value "1". :) Will clarify a bit more.

acioc: Ah, clarification would definitely be good.

changes/2020-07-13_detect-base64-encoded-messages/change.md
MatthewBennington: > decrypt accepts a message as a byte sequence after all

The spec must acknowledge dynamically typed languages though, no?

changes/2020-07-13_detect-base64-encoded-messages/change.md
robin-aws: I was following the example here, which only includes the SHA in "Affected Features" but not in "Affected Specifications": https://raw.githubusercontent.com/awslabs/aws-encryption-sdk-specification/master/changes/2020-05-13_remove-keyring-trace/change.md

@mattsb42-aws Do you think we need to use permalinks in both places? Whatever we decide should be written down in #137.

changes/2020-07-13_detect-base64-encoded-messages/change.md
robin-aws: Absolutely! But it MUST interpret whatever input it is given as a sequence of bytes somehow. If it's idiomatic to accept a string because it quacks like a byte sequence, the operation has to be clear about how it will interpret it as bytes: whether it will use a particular hard-coded encoding like UTF-8, or rely on an encoding attached to the string value, etc. 

In a particular language implementation, I would expect the documentation to cover this if it's not obvious by context.

changes/2020-07-13_detect-base64-encoded-messages/change.md
robin-aws: As per my comment on #137, switching the existing permalink back to relative.

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change should have positive operational impact,
+as it should reduce the number of support engagements with customers
+necessary to help diagnose failures due to this root cause.
```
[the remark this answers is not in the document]
robin-aws: Nice, I like it

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+(currently they are in fact fixed),
+the first two bytes of the Base64 encoding of a valid message are also simple to recognize.
+
+## Drawbacks
+
+This feature technically means the version value
+cannot be incremented beyond the hex value `40`,
+since `41` would then conflict with the Base64-encoded value for `1`.
+This does not seem worth worrying about
+given how rarely we intended to change the version, however.
```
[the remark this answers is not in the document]
robin-aws: To rephrase, it's not that the version cannot advance to 41, it's just that if it does this "SHOULD" becomes impossible to do for all versions, but that doesn't mean it shouldn't be done when possible. I'll improve this.

changes/2020-07-13_detect-base64-encoded-messages/change.md
```diff
@@ -0,0 +1,88 @@
...
+and fail with a more specific error message.
+In particular, the hex values to detect for the current set of versions and types are:
+
+| Version and type (hex) | Base64 encoding (ascii) | Base64 encoding (hex) |
+| ---------------------- | ----------------------- | --------------------- |
+| 01 80                  | A Y ...                 | 41 59 ...             |
+
+Note that the bytes that follow the initial two in the Base64 encoding
+partially depend on subsequent bytes in the binary message format
+and hence are not as predictable.
```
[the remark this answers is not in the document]
robin-aws: I'd like to at least keep the table so that each implementation doesn't have to go through that process themselves.

PR #155, opened 2020-07-13
docs(raw-aes-keyring): raw-aes-keyring MUST fail when given unserializable EC

### LINKED ISSUE #65: Unspecified behavior for RawAESKeyring for invalid input EC
It is possible for the RawAESKeyring to recieve an encryption context as input that is unable to be serialized. Because the RawAESKeyring needs to serialize the EC to use as AAD to the AES Encrypt/Decrypt operation, it is unclear what the correct behavior SHOULD/MUST be in this case.

This gets more complicated in in the OnDecrypt operation, as the EC only needs to be serialized if the keyring attempts decryption of a key.

In our current implementations we have the following behaviors:
- JS: If EC serialization fails while attempting to decrypt a key, continue and do not error
- C: Check if EC can be serialized prior to attempting any decryption (even if input edk list is empty). If it cannot be serialized then immediately return error code.

These implementations do not explicitly violate the spec as written (one can argue that the C one does by returning an error before checking the keys), however are still inconsistent. Update the spec to make it clear what the behavior should be in this case.

MatthewBennington:
_Issue #, if available:_ Resolves #65

_Description of changes:_  Add lines specifying the the raw-aes-keyring MUST fail if it cannot serialize an encryption context.


changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+| ----------------------------------------------------- |
+| [raw-aes-keyring](../../framework/raw-aes-keyring.md) |
+
+## Affected Specifications
+
+This serves as a reference of all specification documents that this change affects.
+
+| Specification                                         |
+| ----------------------------------------------------- |
+| [raw-aes-keyring](../../framework/raw-aes-keyring.md) |
```
acioc, line 20: Should these be SHA's, so we know the version of the spec that this changed?

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+This serves as a reference for all implementations that this change affects.
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
+| Javascript | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-javascript) |
+
+## Definitions
+
+- "serialize", "serialization", and "serializability" all refer to
```
acioc, line 33: np: this probably doesn't need to be a bullet, but up to you

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Specify behavior for RawAESKeyring for invalid input EC
```
acioc, line 4: np: I would write this as Raw AES Keyring instead of RawAESKeyring (since the naming conventions might differ language to language). I would probably also replace all uses of `raw-aes-keyring` with `Raw AES keyring`

I would also fully write out EC, since this is its first occurence.

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Out of Scope
+
+- All keyrings aside from the raw-aes-keyring
+- The encryption context serialization format
+
+## Motivation
+
+The C and Javascript implementations of the raw-aes-keyring check for serializability at different points in OnDecrypt.
+This has potential to lead to different behavior under specific circumstances.
```
acioc, line 61: np: `This has potential to lead to different behavior`-> `This has the potential to lead to different behaviors`

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+which only checks serializability when attempting to decrypt an EDK.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change will improve behavioral consistency between raw-aes-keyring implications.
+This should improve customer experience, and reduce support engagements.
```
acioc, line 77: np: `This should improve customer experience, and reduce support engagements.` -> `This should improve the customer experience and reduce support engagements.`

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Guide-level Explanation
+
+Serialize the encryption/decryption materials' encryption context before attempting encryption.
```
acioc, line 81: This is a bit weird. We say we serialize the decryption material's encryption context before attempting encryption. Should this be encryption/decryption?

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+If serialization fails, the OnEncrypt/OnDecrypt operation MUST fail.
+When encrypting/decryption the datakey, use the already serialized encryption context as the AAD.
+
+## Reference-level Explanation
+
+### OnEncrypt
+
+Before encrypting the plaintext datakey, OnEncrypt MUST attempt to serialize the encryption context.
+If serialization fails, OnEncrypt MUST fail.
+(This is not a change, but is more explicit now)
```
acioc, line 91: np: I would remove the `(`, `)` and add a period. I would also say "this is not a new change" instead of "this is not a change"

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
MatthewBennington: I cribbed the format from other open PRs, which don't seem to do this. Your suggestion makes sense though, seems like we need to better standardize the change doc template.

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+
+### OnEncrypt
+
+Before encrypting the plaintext datakey, OnEncrypt MUST attempt to serialize the encryption context.
+If serialization fails, OnEncrypt MUST fail.
+This is not a new change, but is more explicit now.
+
+### OnDecrypt
+
+Before iterating through the list of EDKs, OnDecrypt MUST attempt to serialize the encryption context.
```
seebees, line 95: Nit: If there are no EDKs in the list that match this name/namspace pair then I don't think the keyring should be required to successfully serialize.
I think that this also means that the spec should be updated.
Because I don't think that
"The keyring MUST perform the following actions on each encrypted data key in the input encrypted data key list, serially, until it successfully decrypts one."
Is correct.
For example it should NOT be attempting to decrypt AWS KMS ciphertext blobs.

MatthewBennington: > Nit: If there are no EDKs in the list that match this name/namspace pair then I don't think the keyring should be required to successfully serialize.

I see the logic behind that (if you look back in the commits on this branch, you'll see that's what I originally wrote). The way I see it though is an unserializable EC means we're in bad enough state that there's no reason to let decryption go on.

> For example it should NOT be attempting to decrypt AWS KMS ciphertext blobs

I believe the spec already accounts for this:

> The keyring MUST attempt to decrypt the encrypted data key if and only if the following is true:
> - ...
> - the key name obtained from the encrypted data key's key provider information has a value equal to this keyring's [key name](#key-name).
> - the key provider ID of the encrypted data key has a value equal to this keyring's [key namespace](#key-namespace).

With the change in #140, the Raw AES Keyring should never attempt to decrypt KMS EDKs.

seebees: The CMM can modify the encryption context.
So if the CMM could modifies the EC in such a way that the Raw keyring would fail.
On the other hand, since we now deny noop keyrings
the raw AES Keyring MUST fail if there are no EDKS
and this strange interaction is better communicated your way.

changes/2020-07-15_specify-behavior-for-raw-aes-keyring-for-invalid-EC/change.md
```diff
@@ -0,0 +1,99 @@
...
+which only checks serializability when attempting to decrypt an EDK.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change will improve behavioral consistency between Raw AES Keyring implications.
+This should improve customer experience and reduce support engagements.
```
acioc, line 77:
```suggestion
This should improve the customer experience and reduce support engagements.
```

PR #157, opened 2020-07-14
feat!: Require generation in the multi-keyring

### LINKED ISSUE #114: Multi keyring generator
Should the multi keyring generator REQUIRE generation?
The specification as written lets you compose multi keyrings that all have generators
in ANY combination.
It is easier to get the organization correct by construction if there is one
and only one path to data key generation.

robin-aws:
Fixes #114

Breaking change since multi-keyrings currently allow a generator to be configured and yet accept encryption materials that already have a plaintext data key defined.


changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,108 @@
...
+# Require Generation in the Multi-Keyring
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                                                                           |
+| ------------------------------------------------------------------------------------------------- |
+| [Multi-Keyring](../../framework/multi-keyring.md)                                                 |
+| [Multi keyring generator](https://github.com/awslabs/aws-encryption-sdk-specification/issues/114) |
```
acioc, line 13: np: we should stick with either `multi-keyring` or `multi keyring`

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,108 @@
...
+
+The multi-keyring has a clearly configured intent of which keyring SHOULD be used to generate the data key,
+but this is not actually enforced during the encryption operation.
+This change strengthens the requirements on multi-keyrings such that if a generator is configured,
+it MUST generate the data key.
+
+## Out of Scope
+
+Detecting at keyring configuration time that encryption will always fail
+because keyrings are not capable of meeting these requirements
```
acioc, line 51: Might be worth saying what "these" requirements are. Is it "because the generator keyring cannot call AWS KMS GenerateDataKey or already has a plaintext data key" or is there something more?

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Require Generation in the Multi-Keyring
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                                                                           |
+| ------------------------------------------------------------------------------------------------- |
+| [Multi-Keyring](../../framework/multi-keyring.md)                                                 |
```
acioc, line 12: Quick extra note. I noticed in another PR you had the specific SHA the feature it changed as opposed to just the link. You should probably use the SHA here and everywhere else multi-keyring.md is linked

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,108 @@
...
+
+If this keyring has a generator keyring,
+this keyring MUST first generate a plaintext data key using the generator keyring.
+If the input encryption materials already include a plaintext data key,
+OnEncrypt MUST fail.
+Otherwise,
+this keyring MUST first call that generator keyring's OnEncrypt
+using the input encryption materials as input.
+If the generator keyring fails OnEncrypt,
+this OnEncrypt MUST also fail.
```
seebees, line 106:
```suggestion
If this keyring has a generator keyring,
and the input encryption materials includes a plaintext data key,
OnEncrypt MUST fail.
This keyring MUST first call that generator keyring's OnEncrypt
using the input encryption materials as input.
If the generator keyring fails OnEncrypt,
this OnEncrypt MUST also fail.
```

changes/2020-07-14_multi-keyring-require-generation/change.md
robin-aws: Agreed, the second was copied from the issue title. Will fix.

changes/2020-07-14_multi-keyring-require-generation/change.md
robin-aws: Left over from an earlier version that had at least one more requirement (that turned out to be already specified). But the general concern is still more broad so I'll indicate that.

changes/2020-07-14_multi-keyring-require-generation/change.md
robin-aws: Agreed, although whether the following section needs it as well is still up in the air: https://github.com/awslabs/aws-encryption-sdk-specification/pull/154#discussion_r455213501

changes/2020-07-14_multi-keyring-require-generation/change.md
robin-aws: Ended up using bullet points to clarify the scoping here instead

robin-aws:
> Looks good! There's no "Reference-level Explanation" (not sure if there needs to be), but that's fine since the reference-level is your spec change anyways.

Yes, in this case there's no detail that would be in one but not the other. Would it be better to label the section as "Guide-level/Reference-level Explanation"?

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,114 @@
...
+that cannot satisfy the requirements of an operation.
+Detecting at keyring configuration time that encryption/decryption will always fail
+because keyrings are not capable of meeting their requirements
+is [tracked separately](https://github.com/awslabs/aws-encryption-sdk-specification/issues/144).
+
+## Motivation
+
+If we do not make a guarantee that the generator will generate,
+then its usefulness is limited.
+Customers may have a compliance requirement that all data keys MUST be generated by a specific HSM,
```
mattsb42-aws, line 63: wording: Avoid "may" when it is not "MAY". In this case I would use "might", "could", or similar.

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,114 @@
...
+between security and usability.
+
+## Security Implications
+
+Removing this flexibility may help identify potential security issues,
+as customers may discover when upgrading their ESDK version
+that data keys are not being generated as intended.
+Correcting this will allow customers
+to improve their security postures
+through improvements such as scoping down permissions.
```
mattsb42-aws, line 86: I would split this into two separate implications:

1. Do the correct thing.
1. Scope down permissions.

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Operational Implications
+
+Customers that are currently using a single multi-keyring instance
+in multiple contexts with inconsistent data key generation patterns
+will encounter errors when upgrading.
+This may lead to an increase in support requests,
+but should also reduce troubleshooting requests
```
mattsb42-aws, line 94:
```suggestion
but SHOULD also reduce troubleshooting requests
```

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,114 @@
...
+Correcting this will allow customers
+to improve their security postures
+through improvements such as scoping down permissions.
+
+## Operational Implications
+
+Customers that are currently using a single multi-keyring instance
+in multiple contexts with inconsistent data key generation patterns
+will encounter errors when upgrading.
+This may lead to an increase in support requests,
```
mattsb42-aws, line 93:
```suggestion
This might lead to an increase in support requests,
```

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -0,0 +1,114 @@
...
+
+- If the input encryption materials already include a plaintext data key,
+  OnEncrypt MUST fail.
+- Otherwise,
+  this keyring MUST first call that generator keyring's OnEncrypt
+  using the input encryption materials as input.
+  If the generator keyring fails OnEncrypt,
+  this OnEncrypt MUST also fail.
+  If the generator keyring returns encryption materials missing a plaintext data key,
+  OnEncrypt MUST fail.
```
mattsb42-aws, line 114: I would split this into individual requirements.

```suggestion
- This keyring MUST first call the generator keyring's OnEncrypt
  using the input encryption materials as input.
- If the generator keyring fails OnEncrypt,
  this OnEncrypt MUST also fail.
- If the generator keyring returns encryption materials missing a plaintext data key,
  OnEncrypt MUST fail.
```

acioc:
> > Looks good! There's no "Reference-level Explanation" (not sure if there needs to be), but that's fine since the reference-level is your spec change anyways.
> 
> Yes, in this case there's no detail that would be in one but not the other. Would it be better to label the section as "Guide-level/Reference-level Explanation"?

I personally prefer renaming - I think renaming makes it clear it wasn't forgotten about and reinforces the standard we're trying to follow.

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -60,10 +60,10 @@ is [tracked separately](https://github.com/awslabs/aws-encryption-sdk-specificat
 
 If we do not make a guarantee that the generator will generate,
 then its usefulness is limited.
-Customers may have a compliance requirement that all data keys MUST be generated by a specific HSM,
-or may want to lock down true least-privilege permissions for their AWS KMS key policies.
+Customers might have a compliance requirement that all data keys MUST be generated by a specific HSM,
+or might want to lock down true least-privilege permissions for their AWS KMS key policies.
 Under the current model,
-customers have to be very careful to make sure that they know
+users have to be very careful to make sure that they know
```
acioc, line 66:
```suggestion
customers have to be very careful to make sure that they know
```

changes/2020-07-14_multi-keyring-require-generation/change.md
```diff
@@ -78,10 +78,10 @@ between security and usability.
 
 ## Security Implications
 
-Removing this flexibility may help identify potential security issues,
-as customers may discover when upgrading their ESDK version
+Removing this flexibility might help identify potential security issues,
+as users might discover when upgrading their ESDK version
 that data keys are not being generated as intended.
```
acioc:
```suggestion
Removing this flexibility might help identify potential security issues.
For example, when upgrading their version of the ESDK, customers might discover
that data keys are not being generated as intended.
```

robin-aws: Thanks, applied this manually

PR #161, opened 2020-07-15
docs(raw-rsa-keyring): Raw RSA Keyring public/private key validation

### LINKED ISSUE #91: Raw RSA keyring public/private key inputs and validation
There are a few points that we need to clarify regarding the raw RSA keyring inputs:

* It MUST accept a public key without a private key.
  * In this case the keyring MUST do nothing on decrypt.
* It MUST accept a private key without a public key.
  * In this case the keyring MUST cause encryption to stop.
* It MUST accept both a private key and a public key.
  * Now captured in #167 ~In this case the keyring SHOULD check to make sure that they are part of the same pair.~
    * ~If the keyring does this check and finds that they are not, it MUST cause initialization to stop.~

MatthewBennington:
_Issue #, if available:_ resolves #91

_Description of changes:_ 

- Require a check during Raw RSA Keyring initialization if two keys are input that they form a matching key pair.
- Clarify language around behavior when OnEncrypt/OnDecrypt is called without a public key/private key respectively.

_Note:_ I went with the keyring "MUST" check that they match, instead of it "SHOULD" check that they match. This is a two-way-door, but I don't see a reason to not check.

### DIFF
--- a/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
+++ b/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
@@ -51,7 +51,7 @@ Clarify language in OnEncrypt and OnDecrypt to require failure, instead of the v
 ## Out of Scope
 
 - All keyrings aside from the Raw RSA Keyring
-- The manner in which the Raw RSA Keyring is build/constructed/instantiated
+- The manner in which the Raw RSA Keyring is initialized
 - How to determine if a pair of keys match
 
 ## Motivation
--- a/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
+++ b/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
@@ -7,19 +7,17 @@
 
 This serves as a reference of all features that this change affects.
 
-| Feature |
-| ------- |
+| Feature                                                                                                                                                   |
+| --------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Raw RSA Keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/31b0534c4259aad365f048b73231545583389c67/framework/raw-rsa-keyring.md) |
-
 
 ## Affected Specifications
 
 This serves as a reference of all specification documents that this change affects.
 
-| Specification |
-| ------------- |
+| Specification                                                                                                                                             |
+| --------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | [Raw RSA Keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/31b0534c4259aad365f048b73231545583389c67/framework/raw-rsa-keyring.md) |
-
 
 ## Affected Implementations
 
@@ -33,7 +31,7 @@ This serves as a reference for all implementations that this change affects.
 ## Definitions
 
 - "match", "matching pair" describe an interoperable RSA public key, and RSA private key
- (i.e. a ciphertext encrypted by the public key can be decrypted by the private key)
+  (i.e. a ciphertext encrypted by the public key can be decrypted by the private key)
 
 ### Conventions used in this document
 
--- a/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
+++ b/changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
@@ -1,0 +1,94 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Raw RSA Keyring public/private key validation
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature |
+| ------- |
+| [Raw RSA Keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/31b0534c4259aad365f048b73231545583389c67/framework/raw-rsa-keyring.md) |
+
+
+## Affected Specifications
+
+This serves as a reference of all specification documents that this change affects.
+
+| Specification |
+| ------------- |
+| [Raw RSA Keyring](https://github.com/awslabs/aws-encryption-sdk-specification/blob/31b0534c4259aad365f048b73231545583389c67/framework/raw-rsa-keyring.md) |
+
+
+## Affected Implementations
+
+This serves as a reference for all implementations that this change affects.
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
+| Javascript | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-javascript) |
+
+## Definitions
+
+- "match", "matching pair" describe an interoperable RSA public key, and RSA private key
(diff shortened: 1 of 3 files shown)


changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
+| Javascript | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-javascript) |
+
+## Definitions
+
+- "match", "matching pair" describe an interoperable RSA public key, and RSA private key
+  (i.e. a ciphertext encrypted by the public key can be decrypted by the private key)
```
acioc, line 34:
```suggestion
"match", "matching pair" describe an interoperable RSA public key and RSA private key
(i.e. a ciphertext encrypted by the public key can be decrypted by the private key)
```

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Summary
+
+Add a requirement to the Raw RSA Keyring that if both the public key and the private key are defined,
+they MUST form a matching pair.
```
acioc, line 47: Is there a way for us to actually validate this without trying to encrypt/decrypt something?

I believe they SHOULD form a matching pair. If they don't, we MUST NOT succeed.

I'm a bit worried that if we say they MUST, there will be an implication that we check that this at keyring initialization time. Do we/ can we?

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+
+Clarify language in OnEncrypt and OnDecrypt to require failure, instead of the vague "MUST NOT succeed".
+
+## Out of Scope
+
+- All keyrings aside from the Raw RSA Keyring
+- The manner in which the Raw RSA Keyring is build/constructed/instantiated
```
acioc, line 54:
```suggestion
- The manner in which the Raw RSA Keyring is initialized
```

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+- How to determine if a pair of keys match
+
+## Motivation
+
+A Raw RSA Keyring with mismatched keys is by construction unable to fulfill its decryption contract.
+As such, we should not allow such a keyring to be constructed.
```
acioc, line 60: I'm not totally sure about these statements. We're saying we cannot allow a keyring to be constructed with mismatched pairs, but we're also saying that determining if they're matched (the way we check this) is out of scope.

I'm worried this will lead to implementation differences (or potentially some languages where we can't do this as part of the keyring initialization process.

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+
+Language clarification around failure is needed,
+as previous language "MUST NOT succeed" could imply a non-failure response is valid
+(such as returning encryption/decryption materials unmodified).
```
acioc, line 64: We can flip this language and say "We MUST only succeed if the keyring was initialized with matching pairs" or something similar and solve this problem without needing to check if the pairs match at initialization.

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
seebees: The problem is that the intention in `key name` is to uniquely identity the wrapping key.
So while it is possible to have have some kind of boundary keyring
that could decrypt with 1 private key only to encrypt with a non-matching pair,
to identify these things as the same makes things confusing.

I will say that if 2 customers are trying to interoperate
with a Raw RSA keyring that they have named the same thing
they MAY want to do _something_ like this.

But in this case, I would suggest that they create 2 keyrings.
One to decrypt and one to encrypt.
If these 2 keyrings have the same `name` values
but different RSA wrapping keys then the requirement is achieved
and the construction is clear.

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+This change will break customers who are using the Raw RSA Keyring with a non-matching keypair.
+However, this is not use case we should encourage.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change SHOULD NOT have any operational implications.
```
seebees, line 77: If it is a breaking change how can it have no operational implications?

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Guide-level Explanation
+
+When you attempt to create a Raw RSA Keyring with both a public and private key,
```
seebees, line 81: Why?  Add some language about how we intend key name/namespace to be used.
I added a comment above about the idea that these values SHOULD
uniquely identify a wrapping key.

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
```diff
@@ -0,0 +1,92 @@
...
+the ESDK MUST check to ensure the keys match.
+If they keys to not match, the creation operation MUST fail.
+
+## Reference-level Explanation
+
+- A check MUST be added in the operation which creates a Raw RSA Keyring
```
seebees, line 87: MUST?  You have SHOULD above?

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
MatthewBennington: > We're saying we cannot allow a keyring to be constructed with mismatched pairs, but we're also saying that determining if they're matched (the way we check this) is out of scope.

The method for determining if keys match seems like an implementation detail to me.

> I'm worried this will lead to implementation differences (or potentially some languages where we can't do this as part of the keyring initialization process).

Are you concerned about different implementations allowing/disallowing certain keypairs? I don't think that's a concern, as long as each implementation checks correctly.

changes/2020-07-15_raw-rsa-keyring-public-private-key-validation/change.md
acioc: From @seebees 's comment above, not every language has a way to do this easily. I'm incredibly worried this will be both burdensome to ESDK developers and be prone to bugs in complex implementations in order to satisfy this requirement.

MatthewBennington: I don't think any runtime has a way to do this easily. If we don't want to require the check, we should have a conversation about whether we want one at all.

PR #163, opened 2020-07-16
feat!: Reserve “aws-crypto-“ in EC input to encrypt

### LINKED ISSUE #23: Define encryption context reserved prefix
Current the spec states that the EC SHOULD reserve any key with the prefix "aws". We should decide on what we want to reserved prefix to be exactly, and update this to a MUST.

Acceptance criteria:
- EC states that a certain prefix MUST be reserved.

robin-aws:
Fixes #23


client-apis/encrypt.md
```diff
@@ -65,7 +65,10 @@ The plaintext to encrypt.
 
 See [encryption context](../framework/structures.md#encryption-context).
 
-The encryption context MUST NOT contain an entry with a [reserved key value](../framework/structures.md#encryption-context).
+The prefix `aws-crypto-` is reserved for internal use by the AWS Encryption SDK;
+see the [the Default CMM spec](default-cmm.md) for one such use.
```
seebees, line 69: I think we should change the default CMM
to error if the provided EC includes `aws-crypto-public-key`.
Since we are removing the MUST from the EC section,
I would just add it there

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
```diff
@@ -0,0 +1,113 @@
...
+involves one or more relevant values related to AWS services.
+This means if any future versions of CMMs packaged with the ESDK
+choose to add additional internal key fields to the encryption context,
+they could be a breaking change.
+
+In addition, stating that the encryption context structure itself
+MUST reserve any set of key fields
+is questionable.
+Interpreted literally, this leads to a contradiction
+since the [Default CMM](../../framework/default-cmm.md) itself adds a key field with the reserved prefix.
```
acioc, line 80:
```suggestion
since the [Default CMM](https://github.com/awslabs/aws-encryption-sdk-specification/blob/2eb5b5fd81a1e8d57655efa3f73573b689e8cd42/framework/default-cmm.md) itself
adds a key field with the reserved prefix.
```

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
```diff
@@ -0,0 +1,113 @@
...
+Therefore, it is clearer to specify this
+as a restriction on user input to the decrypt operation.
+This still allows any [CMM](../../framework/cmm-interface.md) to use the reserved scope.
```
acioc, line 83:
```suggestion
This still allows any [CMM](https://github.com/awslabs/aws-encryption-sdk-specification/blob/2eb5b5fd81a1e8d57655efa3f73573b689e8cd42/framework/cmm-interface.md)
to use the reserved scope.
```

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
```diff
@@ -0,0 +1,113 @@
...
+## Security Implications
+
+By itself this change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+There is still a very slight chance
+this change will break an existing use case.
+This change will be released under a new major version of each implementation,
+and it will at least be possible if difficult
```
acioc, line 103:
```suggestion
and it will at least be possible, if difficult,
```

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
```diff
@@ -0,0 +1,113 @@
...
+for any affected users to migrate to using a different key field.
+
+## Guide-level/Reference-level Explanation
+
+The last two statements in the specification of the [encryption context structure](../../framework/structures.md#encryption-context)
```
acioc, line 108:
```suggestion
The last two statements in the specification of the [encryption context structure](https://github.com/awslabs/aws-encryption-sdk-specification/blob/2eb5b5fd81a1e8d57655efa3f73573b689e8cd42/framework/structures.md#encryption-context)
```

client-apis/encrypt.md
robin-aws: To clarify, we'd specify the reserved prefix in encrypt but ALSO have the default CMM check for this value explicitly, just in case the CMM is used outside of encrypt? I'm fine with that.

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
robin-aws: See #137, ignoring :)

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
robin-aws: See #137, ignoring :)

changes/2020-07-15_encryption-context-reserved-prefix.md/change.md
robin-aws: See #137, ignoring :)

client-apis/encrypt.md
seebees: Yes, I'm saying the default CMM MUST check.
Mostly to keep people safe if they try and pass this value in the EC on encrypt.
(otherwise the default CMM would clobber the value silently)

robin-aws: > Mostly to keep people safe if they try and pass this value in the EC on encrypt.
> (otherwise the default CMM would clobber the value silently)

Just to clarify for the record, people won't be able to pass this value in the EC on encrypt because that is precisely where the `aws-crypto-` prefix will be rejected. But I agree the default CMM should still defend against this independently to guard against this in other potential contexts.

seebees: Yes, better said. 

>to guard against this in other potential contexts.

I think that the primary case is CMM composition. If someone wants to compose and extend the default CMM behavior.

Issue #117, opened 2020-05-26
What crypto materials caches are required?


seebees:
Python and Java have a null cache.
Should everyone?

seebees:
https://github.com/aws/aws-encryption-sdk-c/issues/326

mattsb42-aws:
and the local CMC

alex-chew:
Closed by #166.

Issue #116, opened 2020-05-26
cryptographic materials cache MUST not be unbounded


seebees:
Specify a max number of entries for the cache.

mattsb42-aws:
specifically, this is about the local CMC

alex-chew:
Closed by #166.

PR #166, opened 2020-07-17
docs: refactor CMC and local CMC specs

alex-chew:
_Issue #, if available:_ #1, #116, #117 

_Description of changes:_
> Before this change, the specification of the Cryptographic Materials Cache (CMC) interface gives a only brief explanation of the local CMC:
>
> > The local CMC is a configurable, in-memory, least recently used (LRU) cache.
> > It provides non-blocking, locking, cache entries per cache identifier.
>
> This change captures a stricter set of behaviors upon which existing local CMC implementations already agree, within a specification separate from the abstract CMC interface. These behaviors include how the local CMC bounds its space and time usage, its eviction strategy, and parameters with which users can configure those behaviors.


changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+about the amount of memory they are willing to allocate for cache entries.
+
+We also mandate that the local CMC must accept any non-negative value
+(up to an implementation-defined limit),
+so that in addition to typical use cases (for entry capacity at least 2),
+a user can simulate a null cache (by using entry capacity 0)
+or a "debouncing" cache (by using entry capacity 1).
+(By "debouncing" cache,
+we mean one that caches materials
+upon consecutive cache requests for those materials.)
```
acioc, line 107:
```suggestion
By "debouncing" cache,
we mean one that caches materials
upon consecutive cache requests for those materials.
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+Although these are not necessarily common use cases,
+they are natural generalizations of the local CMC,
+and so without evidence to the contrary
+we have decided that it is worth supporting.
```
acioc, line 111:
```suggestion
and so without evidence to the contrary,
we have decided that they are worth supporting.
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+Because the null cache is a special case of the local CMC as defined above,
+it is also unnecessary to individually specify the null CMC as its own entity,
+which in turn reduces specification and implementation effort.
+
+Continuing with the theme of avoiding unbounded caches,
+we note that it is dangerous for cryptographic materials to remain in the cache
+for longer than necessary,
+since doing so increases the temporal attack surface.
+This change therefore mandates that
+the local CMC associate a time-to-live (TTL) control to each cache entry,
```
acioc, line 121:
```suggestion
the local CMC associates a time-to-live (TTL) control to each cache entry,
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Eviction
+
+With the above rules in place to dictate how long and in what quantity
+the local CMC can retain cache entries,
+we turn our attention to when and how the local CMC evicts entries.
+
+This change mandates that the local CMC evicts some TTL-expired entries
+on every Get Cache Entry and Put Cache Entry operation,
+and also that the it evicts any additional entries as necessary
```
acioc, line 140:
```suggestion
and also that it evicts any additional entries as necessary
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Drawbacks
+
+Mandating that the local CMC support entry capacity values of 0 and 1,
+may result in edge cases in LRU removal logic
+(as hinted in the [C implementation](https://github.com/aws/aws-encryption-sdk-c/blob/aa85ca224d550cfe110e2112821a84506b9aca3e/source/local_cache.c#L924)).
+However, we believe that the functional benefits
+outweigh the complexity of the edge cases.
+
+By allowing the local CMC not to periodically evict TTL-expired entries,
```
acioc, line 182:
```suggestion
By not requiring the local CMC to periodically evict TTL-expired entries,
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change will break any customer
+who relies on the local CMC behavior in the AWS Encryption SDK for C that,
+if the customer attempt to set an entry capacity of less than 2,
+instead sets the entry capacity to 2.
```
acioc, line 199:
```suggestion
who relies on the local CMC behavior in the AWS Encryption SDK for C,
if the customer attempt to set an entry capacity not equal to 2.
```
?

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Guide-level Explanation
+
+The AWS Encryption SDK provides a build-in local Cryptographic Materials Cache (local CMC),
+which implements the CMC interface.
+The local CMC is an in-memory, least recently used (LRU) cache.
+It provides non-blocking, locking, cache entries per cache identifier.
```
acioc, line 206: Bit confused here. Are they non-blocking or are they locking?

framework/local-cryptographic-materials-cache.md
```diff
@@ -0,0 +1,104 @@
...
+is the number of least recently used entries that the local CMC
+MUST check during [pruning](#pruning)
+for TTL-expired entries to evict.
+
+## Behaviors
+
+### Put Cache Entry
+
+When calling the Put Cache Entry operation,
+the user MUST provide a time-to-live (TTL) value.
```
acioc, line 78: So, this goes a bit against `a point in time` from above.

Are customers providing a time delta (like 5 seconds) or are they providing a timestamp for when they want the element to expire. I would recommend this value specifically be a delta, so we don't have to worry about different timedate formats, time zones, etc.

Internally, we can store it however we want, but we should have customers provide the difference in time, not "a point in time" (a timestamp) like we define TTL higher up.

alex-chew: I intentionally define TTL abstractly as "represent[ing] a point in time", rather than as "a time delta" or "a timestamp", so that implementations have the freedom to choose both their internal representation and the representation of the TTL parameter in Put Cache Entry.

I'm hesitant to _require_ that implementations accept the TTL parameter as a time delta _and not_ as a timestamp, since I think individual implementations are in the best position to decide what's right for them.

changes/2020-07-14_refactor-cmc-spec/change.md
alex-chew: They are both. Each entry is guarded by a lock, but if an actor attempts to acquire a locked entry, the attempt immediately fails and returns execution to the actor (rather than blocking execution until the entry is released and the attempt succeeds).

changes/2020-07-14_refactor-cmc-spec/change.md
acioc: I'm not super opinionated on either point. I can see what you're saying though, so totally up to you.

changes/2020-07-14_refactor-cmc-spec/change.md
acioc: What about just...

"who relies on the local CMC behavior in the AWS Encryption SDK for C
if the customer attempt to set an entry capacity less than 2"

?

changes/2020-07-14_refactor-cmc-spec/change.md
acioc: Interesting - that's fine, but I would recommend clarifying this behavior before the first use.

framework/local-cryptographic-materials-cache.md
acioc: I pretty strongly think accepting timestamps will lead to implementation errors or potential customer confusion, but I'll leave this up to you/the rest of the team.

changes/2020-07-14_refactor-cmc-spec/change.md
alex-chew: Gotcha, going to resolve this then.

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+
+The change also recommends that the local CMC evict TTL-expired entries on a periodic basis,
+if the implementation's programming language and runtime make doing so feasible.
+By doing so,
+the local CMC ensures that no entry lives for more than a constant amount of time past its TTL;
+without periodic eviction,
+a TTL-expired entry could remain in the cache indefinitely
+if no user makes any Get Cache Entry or Put Cache Entry call
+after the entry enters the cache.
+The change does not mandate period eviction because
```
seebees, line 155:
```suggestion
The change does not mandate periodic eviction because
```

changes/2020-07-14_refactor-cmc-spec/change.md
```diff
@@ -0,0 +1,247 @@
...
+concurrency is complex to implement in some programming languages and runtimes.
```
seebees, line 156:
```suggestion
concurrency is complex to implement.
```

changes/2020-07-14_refactor-cmc-spec/change.md
seebees: Just to be super clear, there is no locking in JS.
Or, this kind of operation is atomic.

alex-chew: I've updated the wording to use "atomic" for clarity.

framework/local-cryptographic-materials-cache.md
alex-chew: That's fair. I'll change the wording to explain that the caller provide a time delta specifically, whereas the local CMC can choose how to derive the TTL value.

PR #177, opened 2020-07-20
docs: add AES example and rules for example templates

mattsb42-aws:
_Issue #, if available:_

#143 

_Description of changes:_

Add AES example and example template rules.

I don't *think* that adding the example template rules necessitate a change doc as I believe that they are covered under the intend of the examples change doc. I found them necessary to add in order to frame a template, and including a template in the change with them was useful both for me to clarify them for myself and to demonstrate how each section works.

### DIFF
--- a/examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
+++ b/examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
@@ -1,0 +1,132 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Raw AES keyring example
+
+Implementations of this example MUST follow the rules defined in
+[Example Templates](../../../examples.md#example-templates).
+
+## Implementations
+
+- [Python (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/src/keyring/raw_aes/raw_aes.py)
+- [Python Master Key Provider (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/src/master_key_provider/raw_aes/raw_aes.py)
+- [Java (DEV)](https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/examples/java/com/amazonaws/crypto/examples/keyring/rawaes/RawAes.java)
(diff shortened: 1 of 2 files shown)


examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
```diff
@@ -0,0 +1,132 @@
...
+
+Implementations of this example MUST follow the rules defined in
+[Example Templates](../../../examples.md#example-templates).
+
+## Implementations
+
+- [Python (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/src/keyring/raw_aes/raw_aes.py)
+- [Python Master Key Provider (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/src/master_key_provider/raw_aes/raw_aes.py)
+- [Java (DEV)](https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/examples/java/com/amazonaws/crypto/examples/keyring/rawaes/RawAes.java)
+- [Java Master Key Provider (DEV)](https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/examples/java/com/amazonaws/crypto/examples/masterkeyprovider/rawaes/RawAes.java)
```
acioc, line 14: np: Should this be a table instead?

Also, what about JavaScript and C?

examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
```diff
@@ -0,0 +1,132 @@
...
+       key_name="my AES wrapping key",
+       wrapping_key=key,
+   )
+   ```
+
+1. Encrypt plaintext data.
+
+   ```python
+   # Encrypt your plaintext data.
+   ciphertext, _encrypt_header = aws_encryption_sdk.encrypt(
```
acioc, line 94:
```suggestion
   ciphertext, encrypt_header = aws_encryption_sdk.encrypt(
```

I wasn't sure if the `_` was supposed to be there.

examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
mattsb42-aws: Neither JS nor C have these examples yet. We use a simple list for similar things elsewhere; IMO if we only have one column it's not worth a table.

examples/templates/configuration/raw-keyrings/raw-aes-keyring.md
mattsb42-aws: It is, yes. In Python in this use, that denotes "I need to collect this variable and I want to give it a name for clarity, but I do not intend to ever use it."

PR #179, opened 2020-07-20
docs: add API usage examples templates

mattsb42-aws:
_Issue #, if available:_

#143 

_Description of changes:_

**NOTE: The specific Python code in these templates will change before keyrings are released. I suggest that either we hold this in limbo until those changes are finalized or we update the code in the template when we come back to update links from the keyring branch to the master branch.**

### DIFF
--- a/examples/templates/api/change-algorithm-suite.md
+++ b/examples/templates/api/change-algorithm-suite.md
@@ -8,8 +8,8 @@ Implementations of this example MUST follow the rules defined in
 
 ## Implementations
 
-- [Python (DEV)]()
-- [Java (DEV)]()
+- [Python (DEV)](https://github.com/aws/aws-encryption-sdk-python/blob/keyring/examples/src/onestep_unsigned.py)
+- [Java (DEV)](https://github.com/aws/aws-encryption-sdk-java/blob/keyring/src/examples/java/com/amazonaws/crypto/examples/OneStepUnsigned.java)
 
 ## Definitions
 
--- a/examples/templates/api/change-algorithm-suite.md
+++ b/examples/templates/api/change-algorithm-suite.md
@@ -1,0 +1,133 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Changing the algorithm suite example
+
+Implementations of this example MUST follow the rules defined in
+[Example Templates](../../../examples.md#example-templates).
+
+## Implementations
+
+- [Python (DEV)]()
+- [Java (DEV)]()
+
+## Definitions
+
+### Conventions used in this document
+
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Header
+
+```python
+# This example shows how to specify an algorithm suite
+# when using the one-step encrypt and decrypt APIs.
+#
+# In this example, we use an AWS KMS customer master key (CMK),
+# but you can use other key management options with the AWS Encryption SDK.
+# For examples that demonstrate how to use other key management configurations,
+# see the ``keyring`` and ``master_key_provider`` directories.
+#
+# The default algorithm suite includes a message-level signature
+# that protects you from an attacker who has *decrypt* but not *encrypt* capability
+# for a wrapping key that you used when encrypting a message
+# under multiple wrapping keys.
+#
+# However, if all of your readers and writers have the same permissions,
+# then this additional protection does not always add value.
+# This example shows you how to select another algorithm suite
+# that has all of the other properties of the default suite
+# but does not include a message-level signature.
+```
+
+## Summary
+
+```python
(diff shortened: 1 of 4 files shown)


examples/templates/api/change-algorithm-suite.md
```diff
@@ -0,0 +1,133 @@
...
+## Header
+
+```python
+# This example shows how to specify an algorithm suite
+# when using the one-step encrypt and decrypt APIs.
+#
+# In this example, we use an AWS KMS customer master key (CMK),
+# but you can use other key management options with the AWS Encryption SDK.
+# For examples that demonstrate how to use other key management configurations,
+# see the ``keyring`` and ``master_key_provider`` directories.
```
acioc, line 33: Do we have these directories in the spec/will we? If not, should we link to something else here?

This applies to this doc page and others in this PR.

examples/templates/api/change-algorithm-suite.md
```diff
@@ -0,0 +1,133 @@
...
+```python
+# Demonstrate requesting a specific algorithm suite through the one-step encrypt/decrypt APIs.
+```
+
+## Inputs
+
+- **source plaintext** :
+  Plaintext to encrypt
+- **AWS KMS CMK** :
+  The ARN of an AWS KMS CMK that protects data keys
```
acioc, line 58: Might be good to link to AWS docs (for ARN)? Either way, might also be good to note if this can be an alias ARN.

examples/templates/api/change-algorithm-suite.md
```diff
@@ -0,0 +1,133 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Changing the algorithm suite example
+
+Implementations of this example MUST follow the rules defined in
+[Example Templates](../../../examples.md#example-templates).
```
acioc, line 7: This applies here and in other documents in this PR.

I think this path isn't resolving properly.

acioc: I think these might just need to be 

[Example Templates](../../examples.md#example-templates).

mattsb42-aws: doh; that is correct; I copy-pasted that from the raw keyring examples and they have at least one extra directory layer x_x

examples/templates/api/change-algorithm-suite.md
seebees: Also, for `master_key_provider` not all implementations support this.
We may want to try and remove it?

examples/templates/api/change-algorithm-suite.md
```diff
@@ -0,0 +1,133 @@
...
+```python
+# This example shows how to specify an algorithm suite
+# when using the one-step encrypt and decrypt APIs.
+#
+# In this example, we use an AWS KMS customer master key (CMK),
+# but you can use other key management options with the AWS Encryption SDK.
+# For examples that demonstrate how to use other key management configurations,
+# see the ``keyring`` and ``master_key_provider`` directories.
+#
+# The default algorithm suite includes a message-level signature
```
seebees, line 35: If the "default algorithm suite" ever changes, how are we going to find all these places to update them?

examples/templates/api/change-algorithm-suite.md
seebees: Or a link to the KMS examples?
We don't want any given example
to be so complete that it tries to explain everything.

examples/templates/api/change-algorithm-suite.md
```diff
@@ -0,0 +1,133 @@
...
+
+   ```python
+   # Demonstrate that the ciphertext and plaintext are different.
+   assert ciphertext != source_plaintext
+   ```
+
+1. Decrypt encrypted data.
+
+   ```python
+   # Decrypt your encrypted data using the same keyring you used on encrypt.
```
seebees, line 107: Some of this language is getting copy/paste.
Not a blocker, but in the long run, some more tokenization may be in order.

examples/templates/api/streaming-file-to-file.md
```diff
@@ -0,0 +1,138 @@
...
+   ```python
+   # Create the keyring that determines how your data keys are protected.
+   keyring = AwsKmsKeyring(generator_key_id=aws_kms_cmk)
+   ```
+
+1. Encrypt plaintext data.
+
+   ```python
+   # Open the files you want to work with.
+   with open(source_plaintext_filename, "rb") as plaintext, open(ciphertext_filename, "wb") as ciphertext:
```
seebees, line 88: Yup, I think that this is so language specific,
that maybe nothing is better?
(in the long run :) no need to block )

examples/templates/api/streaming-in-memory.md
```diff
@@ -0,0 +1,131 @@
...
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Header
+
+```python
+# This example shows how to use the streaming encrypt and decrypt APIs on data in memory.
```
seebees, line 27: I like the idea here,
but maybe a little more complicated example?
Like why are we doing this?
Is the idea like a proxy where
you take a network socket and decrypt
it on the fly and send it to some back end server?

PR #210, opened 2020-07-22
docs: fix nitpicks in KMS Keyring spec

alex-chew:
_Issue #, if available:_ N/A

_Description of changes:_ Fix a couple typos, and change a "MAY define" to "MUST optionally define".

### DIFF
--- a/framework/kms-keyring.md
+++ b/framework/kms-keyring.md
@@ -74,7 +74,7 @@ See [AWS Documentation](https://docs.aws.amazon.com/kms/latest/APIReference/API_
 
 An AWS KMS API for decrypting ciphertext previously encrypted by [GenerateDataKey](#aws-kms-generatedatakey) or [Encrypt](#aws-kms-encrypt).
 
-See [AWS Documenetation](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html).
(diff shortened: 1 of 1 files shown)


framework/kms-keyring.md
```diff
@@ -74,15 +74,15 @@ See [AWS Documentation](https://docs.aws.amazon.com/kms/latest/APIReference/API_
...
+See [AWS Documentation](https://docs.aws.amazon.com/kms/latest/APIReference/API_Decrypt.html).
 
 ## Inputs
 
 On keyring initialization, a keyring MUST define the following:
 
 - [Client Supplier](#client-supplier)
 
-On keyring initialization, a keyring MAY define the following:
+On keyring initialization, a keyring MUST optionally define the following:
```
lavaleri, line 85: What about "MUST accept the following as optional inputs"?

In which case I'd also edit the above to say "a keyring MUST accept the following as required inputs:"

lavaleri: We could also address this in a separate PR which standardizes across all keyrings, and merge this in as is.

lavaleri: Addressing concern in https://github.com/awslabs/aws-encryption-sdk-specification/pull/213

PR #220, opened 2020-07-29
Add proposal for adding Keyring CMM and updating Default CMM definition

lavaleri:
_Issue #, if available:_ Proposal for reframing the Default CMM

_Description of changes:_

### DIFF
--- a/proposals/2020-07-28_keyring-cmm/proposal.md
+++ b/proposals/2020-07-28_keyring-cmm/proposal.md
@@ -1,0 +1,199 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Base CMM
+
+## Affected Features
+
+This serves as a reference of all features that this change affects.
+
+| Feature                                       |
+| --------------------------------------------- |
+| [Default CMM](../../framework/default-cmm.md) |
+
+## Affected Specifications
+
+This serves as a reference of all specification documents that this change affects.
+
+| Specification                                 |
+| --------------------------------------------- |
+| [Default CMM](../../framework/default-cmm.md) |
+
+## Affected Implementations
+
+This serves as a reference for all implementations that this change affects.
+
+| Language   | Repository                                                                            |
+| ---------- | ------------------------------------------------------------------------------------- |
+| Python     | [aws-encryption-sdk-python](https://github.com/aws/aws-encryption-sdk-python)         |
+| Java       | [aws-encryption-sdk-java](https://github.com/aws/aws-encryption-sdk-java)             |
+| C          | [aws-encryption-sdk-c](https://github.com/aws/aws-encryption-sdk-c)                   |
+| Javascript | [aws-encryption-sdk-javascript](https://github.com/aws/aws-encryption-sdk-javascript) |
+
+## Definitions
+
+### Conventions used in this document
+
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Summary
+
+The Default CMM is an implementation of the CMM interface that serves a majority of use cases,
+using [keyrings](../../framework/keyring-interface.md) or
+[master key providers](../../framework/master-key-provider-interface.md) to get and
+decrypt materials.
+
+This change proposes that we instead define that implementation under new names,
+Keyring CMM and Master Key Provider CMM,
+and define the Default CMM to be the specific composition of one or more CMMs
+that customers SHOULD use.
+
+## Out of Scope
+
+- The behavior of any existing CMMs implementations.
+
+## Motivation
+
+Prior to this change, the Default CMM serves two purposes.
+First, it provides a base implementation of a CMM that uses a
+keyring or master key provider to ensure valid materials.
+Second, it serves as safe default that can be used for most use cases.
+
+While there is currently no conflict in these purposes,
+this might not always be true.
+In the future we might want to update the behavior customers get when they choose some "default"
+CMM option.
+
+One way to address this would be to directly add the new desired behavior
+to the current Default CMM implementation,
+however this makes our provided CMMs less composable.
+Keeping the CMM composable is important because it allows users to assert desired properties
+[by construction](../../tenets.md#correct-by-construction).
+Because it is important to retain this base implementation,
+we should define a CMM whose sole purpose is to be this base implementation.
+We will call this CMM implementation the Keyring CMM.
+For implementations which need to support master key provider, we will also define a
+Master Key Provider CMM.
+
+Similar to how we define a set of algorithm suites and define a default algorithm suite,
+we should similarly have a set of provided CMM implementations
+and a Default CMM which represents a particular configuration of one of those CMMs
+(possible composed with other CMMs).
+Right now, we define that the Default CMM is "the Keyring CMM".
+Just as an example, if we wanted to provide caching as a default behavior,
+the Default CMM could be defined as "the Caching CMM configured with a Keyring CMM."
+By defining the default in this way, we have greater flexibility to update the default
+CMM that customers use in the future without losing the composability of these implementations.
+
+Such a default MUST NOT allow any options other than the keyring to be configured in order
+to allow us to more easily maintain backwards compatability with possible Default CMM updates.
+
+## Drawbacks
+
+This change SHOULD NOT have any drawbacks.
+
+## Security Implications
+
+This change SHOULD NOT have any security implications.
+
+## Operational Implications
+
+This change SHOULD NOT have any operational implications.
+
+This change MUST maintain behavior for customers who are using a Default CMM.
+
+## Guide-level Explanation
+
+This change defines the [Keyring CMM](#keyring-cmm) and
+the [Master Key Provider CMM](#master-key-provider-cmm)
+as base CMM implementations that can be used "as is" or composed with other CMM.
+
+This change also defines a [Default CMM](#default-cmm) as a specific composition of one or more
+CMMs that the AWS Encryption SDK provides as a safe default that can be used for most use cases.
+
+Customers SHOULD either use the [Default CMM](#default-cmm)
+or compose a CMM using the AWS Encryption SDK's provided CMM implementations
+(including the Keyring CMM).
+
+### Default CMM
+
+The Default CMM is a specific composition and configuration of one or more
+CMM implementations provided by the AWS Encryption SDK.
+
+The specific CMM composition describes a safe default that serves most use cases.
+
+The CMM composition defined by the Default CMM is the [Keyring CMM](#keyring-cmm).
+
+As the AWS Encryption SDK provides more CMMs implementations that would benefit default use cases,
+we expect to update the composition defined by Default CMM to include these new CMMs.
+Any update to the composition defined by Default CMM SHOULD be backwards compatable.
+
+To ensure that its use is correct by construction,
+the instantiation of the Default CMM MUST NOT take any options
+except for the keyring that should be composed with the Default CMM.
+
+Functionality exposed by the AWS Encryption SDK MUST NOT require the Default CMM.
+
+### Keyring CMM
+
+The Keyring CMM is a base CMM implementation that
+[gets encryption materials](../../framework/cmm-interface.md#get-encryption-materials)
+and [decrypts materials](../../framework/cmm-interface.md#decrypt-materials)
+through use of a [keyring](../../framework/keyring-interface.md) configured on instantiation.
+
+This CMM can be used as is or composed with other CMMs,
+but is also provided as the [Default CMM](#default-cmm).
+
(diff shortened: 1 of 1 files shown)


proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,198 @@
...
+Any desired CMM implementation that intends to use master key providers to ensure valid materials
+SHOULD compose with the Master Key Provider CMM.
+
+## Reference-level Explanation
+
+### Default CMM
+
+On initialization it MUST accept:
+
+- an underlying [Keyring](../../framework/keyring-interface.md)
```
seebees, line 176: What about a master key provider?

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,198 @@
...
+(including the Keyring CMM).
+
+### Default CMM
+
+The Default CMM is a specific configuration of a
+CMM implementation provided by the AWS Encryption SDK.
+
+The specific CMM configuration describes a safe default that serves most use cases.
+
+The CMM configuration defined by the Default CMM is the [Keyring CMM](#keyring-cmm) as is
```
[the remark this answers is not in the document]
lavaleri: I'm not sure I understand. If you are suggesting that we be able to initialize the KeyringCMM with a "default alg" option that overrides the default defined in the alg suite, then that would be changing the current behavior of the Default CMM and is out of scope.

proposals/2020-07-28_keyring-cmm/proposal.md
lavaleri: This one makes me sad, but you're right.

proposals/2020-07-28_keyring-cmm/proposal.md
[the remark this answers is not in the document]
seebees: I was suggesting that the Keyring CMM takes an option for "default alg",
and then when the Keyring CMM is created in the Default CMM,
then the option would be what we define as a default.

So the ESDK default algorithm is defined in the Default CMM and not in the Keyring CMM.

proposals/2020-07-28_keyring-cmm/proposal.md
mattsb42-aws: No, we want MKP to be a different CMM.

proposals/2020-07-28_keyring-cmm/proposal.md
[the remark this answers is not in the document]
mattsb42-aws: What if the keyring CMM just requires that the request identify the algorithm suite?

I'm loathe to squeeze opinions about algorithm suite into a CMM that SHOULD only care about wrangling keyrings.

proposals/2020-07-28_keyring-cmm/proposal.md
lavaleri: This would break backwards compatability for users who are already using a DefaultCMM constructor that takes in a MKP. It would force those users to update their code to get it to work without actually providing any benefit to them.

But perhaps it's worth it now in order for this interface to be strictly Keyrings?

My other concern is that if we do add additional behaviors to the default CMM, would we want those changes to only people who are using Keyrings? Even if MKPs are deprecated, I could see value in being able to update their DefaultCMM behavior in a non-breaking way.

proposals/2020-07-28_keyring-cmm/proposal.md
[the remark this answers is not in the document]
lavaleri: I understand what you're proposing now. This would mean that the Default CMM is no longer *just* a configuration. Unless we also want to move to the "default to the default alg CMM" logic out of the Keyring CMM and into a new CMM.

If we were to do something like this, I agree with Matt's approach, the KeyringCMM should just fail if it doesn't get an alg suite. I'm not sure if splitting up the logic this way is useful or not though.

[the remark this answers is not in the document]
seebees: This makes sense, but the keyring CMM should be easy to use on its own.
If customers chose to use _only_ the keyring CMM and inject that into the ESDK, it would not work in the default case.
Maybe that is a good thing though.

What I would like is that the default algorithm suite for the ESDK is defined in the Default CMM, not the keyring CMM.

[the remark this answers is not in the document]
mattsb42-aws: > What I would like is that the default algorithm suite for the ESDK is defined in the Default CMM, not the keyring CMM.

Agreed. If you're opting out of the default then you don't get any of the defaults. I think this is reasonable.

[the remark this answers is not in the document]
seebees: For clarity, we do not want to have the Keyring CMM take a REQUIRED option "default alg"?

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+
+The Default CMM is a specific configuration of a
+CMM implementation provided by the AWS Encryption SDK.
+
+The specific CMM configuration describes a
+[safe default](https://github.com/awslabs/aws-encryption-sdk-specification/blob/2ec7674c304c408c2a32d412e834939c73f68d80/tenets.md#sensible-defaults)
+that serves most use cases.
+
+The CMM configuration defined by the Default CMM is the [Keyring CMM](#keyring-cmm) as is
+(The Keyring CMM provides no additional options other than specifying an underlying Keyring).
```
acioc, line 137:
```suggestion
(the Keyring CMM provides no additional options other than specifying an underlying keyring).
```

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+
+As the AWS Encryption SDK provides more CMM implementations that would benefit default use cases,
+we expect to update the Default CMM to a configuration which composes such CMMs together provide
```
acioc, line 140:
```suggestion
we expect to update the Default CMM to a configuration which composes such CMMs together in order to provide
```

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+useful properties.
+Any update to the composition defined by Default CMM SHOULD be backwards compatable.
+
+To ensure that its use is correct by construction,
+the instantiation of the Default CMM MUST NOT take any options
```
acioc, line 145:
```suggestion
the instantiation of the Default CMM MUST NOT take any options,
```

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+SHOULD compose with the Master Key Provider CMM.
+
+## Reference-level Explanation
+
+### Default CMM
+
+On initialization, the caller MUST provide exactly one of the following:
+
+- [Keyring](../../framework/keyring-interface.md)
+- If this AWS Encryption SDK implementations provides a [Master Key Provider CMM](#master-key-provider-cmm),
```
acioc, line 184:
```suggestion
- If this AWS Encryption SDK implementation provides a [Master Key Provider CMM](#master-key-provider-cmm),
```

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Keyring CMM
+
+The specification of the Keyring CMM MUST be the specification of the
+Default CMM prior to this change.
+
+### Master Key Provider CMM
+
+The specification of the Master Key Provider CMM MUST be the specification of
+the Default CMM, except describing use of a master key provider instead of a keyring
```
acioc, line 208:
```suggestion
The specification of the Master Key Provider CMM MUST be the specification of the Default CMM,
except describing the use of a master key provider instead of a keyring
```

proposals/2020-07-28_keyring-cmm/proposal.md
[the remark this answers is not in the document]
acioc: I might be misunderstanding this conversation, but near the end of the doc, we say "This CMM MUST NOT offer any additional features beyond the composed CMM created above." To me, this means the default CMM MAY offer the same configuration options as the thing it is configuring. However, it MAY additionally support sane defaults.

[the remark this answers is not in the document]
lavaleri: After offline discussion:

- It is important that the Default CMM MUST represent an expression of default values without accompanying logic
  - This defines all defaults for encryption and decryption in one place, and additionally makes these defaults more extendable for any future code that may use CMMs but is not necessarily our current Encrypt/Decrypt client APIs. 
  - By ensuring this is a strict configuration, we ensure that each behavior is encapsulated in a composable CMM which can be reused by customers, making it easy to compose CMMs which opt out of certain default behaviors or reuse default behaviors within another configuration.
  - This also makes updating the default behaviors easy, and able to be done in a way to preserves composability of our behaviors.
- As such, the default algorithm suite is defined by the Default CMM
  - Customers MUST be able to get this default algorithm suite statically
- One more behavior of the KeyringCMM in this version of the spec is that it defaults to the default algorithm. This is not desired and this behavior should be encapsulated outside the Keyring CMM implementation.
  - The Keyring CMM MUST require an algorithm suite
- We should add a new CMM implementation DefaultRequestValues (name TBD) that is solely responsible for setting default values in the CMM's materials requests, and then passes that request to an underlying CMM.
  - For this proposal, this should ONLY be setting some default algorithm suite for the get encryption materials request.
- The Default CMM should thus represent the composition of this new DefaultRequestValues CMM and a Keyring CMM.
- We need to add examples for composing CMMs together.

I will update the doc with this motivation and behavior.

proposals/2020-07-28_keyring-cmm/proposal.md
lavaleri: I prefer without a comma here, but feel free to tell me I'm wrong.

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,209 @@
...
+
+This change SHOULD NOT have any operational implications.
+
+This change MUST maintain behavior for customers who are using a Default CMM.
+
+## Guide-level Explanation
+
+This change defines the [Keyring CMM](#keyring-cmm) and
+the [Master Key Provider CMM](#master-key-provider-cmm)
+as base CMM implementations that can be used "as is" or composed with other CMM.
```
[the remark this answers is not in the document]
lavaleri: went with other CMMs

proposals/2020-07-28_keyring-cmm/proposal.md
mattsb42-aws: To clarify: We want the keyring CMM to only handle keyrings and the MKP CMM to only handle CMMs. The DefaultCMM is responsible for determining which one to use based on the input.

lavaleri:
I don't like the name "Set Unspecified Request Values CMM", but I could not think of any other name that didn't use "Default" in some way that could be interpreted wrongly. Still trying to brainstorm, but accepting recommendations!

proposals/2020-07-28_keyring-cmm/proposal.md
```diff
@@ -0,0 +1,293 @@
...
+[gets encryption materials](../../framework/cmm-interface.md#get-encryption-materials)
+and [decrypts materials](../../framework/cmm-interface.md#decrypt-materials)
+through use of a [keyring](../../framework/keyring-interface.md) configured on instantiation.
+
+It requires an algorithm suite to be set in the Get Encryption Materials Request.
+
+Any desired CMM implementation that intends to use keyrings to ensure valid materials SHOULD
+compose with the Keyring CMM.
+
+### Set Unspecified Request Values CMM
```
robin-aws, line 184: Could we have the Encrypt operation fill in the default when not provided as input, instead of it being the CMM's responsibility?

lavaleri: We have considered this (and this is what Java/Python originally did, I believe), but we want this to be defined in the CMM for two reasons:
- The logic for everything the ESDK expresses as a "default" is encapsulated at the same level, in the Default CMM that we provide to the user. This is simple, and means that we have only one place in code where we define what "default" is.
- If this logic is in the CMM as opposed to the client level, then any other code that theoretically can be built on top of CMMs is able to grab the defaults, instead of needing to duplicate the "set some default for the algorithm suite" behavior.

robin-aws: On thinking about this further, I'm not really convinced by either reason:

* "in the Encrypt operation" would be an equally valid single place to encapsulate the defaults
* Setting a default value doesn't seem to be difficult enough to worry about never duplicating IF there is significant code out there using CMMs outside the context of Encrypt/Decrypt

I agree the "Set Unspecified Request Values CMM" is valid and possible, but it feels like a really heavyweight solution. It would be a lot simpler to say that CMMs require the algorithm suite to be set in all cases.

robin-aws: After offline discussion I'm at least convinced this is a reasonable solution - I'll make suggestions in the text for how we can justify it a bit more strongly.

robin-aws:
> I don't like the name "Set Unspecified Request Values CMM", but I could not think of any other name that didn't use "Default" in some way that could be interpreted wrongly. Still trying to brainstorm, but accepting recommendations!

How about "Default-Providing CMM"?

PR #223, opened 2020-08-05
chore: Reserve algorithm suite id

seebees:
`00 00` should be reserved for extension or other uses.

### DIFF
--- a/framework/algorithm-suites.md
+++ b/framework/algorithm-suites.md
@@ -101,7 +101,10 @@ ECDSA-Sig-Value ::= SEQUENCE {
 ## Supported Algorithm Suites
 
 The following table includes the algorithm suites supported by the AWS Encryption SDK.
-The values for each field of each algorithm suite is specified below. The Algorithm Suite ID MUST NOT be `00 00`.
+The values for each field of each algorithm suite is specified below. 
+The value `00 00` is reserved
+and MUST NOT be used
+as an Algorithm Suite ID in the future.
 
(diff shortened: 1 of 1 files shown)


framework/algorithm-suites.md
```diff
@@ -101,7 +101,7 @@ ECDSA-Sig-Value ::= SEQUENCE {
 ## Supported Algorithm Suites
 
 The following table includes the algorithm suites supported by the AWS Encryption SDK.
-The values for each field of each algorithm suite is specified below.
+The values for each field of each algorithm suite is specified below. The Algorithm Suite ID MUST NOT be `00 00`.
```
lavaleri, line 104:
```suggestion
The values for each field of each algorithm suite is specified below.
The Algorithm Suite ID MUST NOT be `00 00`.
```

robin-aws: I'd phrase it differently. Currently the Algorithm Suite ID MUST NOT be ANY value other than those in the table. How about:

```suggestion
The values for each field of each algorithm suite is specified below. 
The Algorithm Suite ID `00 00` is reserved for future use and MUST NOT be added in the future.
```

seebees: Let me get in on the game.
`00 00` can not both be an ID and not be added.
But @robin-aws I like thrust.
How about?

```suggestion
The values for each field of each algorithm suite is specified below. 
The value `00 00` is reserved
and MUST NOT be used
as an Algorithm Suite ID in the future.
```

robin-aws: Even better :)

PR #224, opened 2020-08-05
fix: apply consistent behavior for refusing to write or fullfil any decryption contract

mattsb42-aws:
_Issue #, if available:_

These are changes I missed making when defining decryption contracts and keyring failure modes. Now that no-ops are not allowed, if a keyring would have previously been a no-op, it MUST now fail.

_Description of changes:_

### DIFF
--- a/framework/keyring-interface.md
+++ b/framework/keyring-interface.md
@@ -5,9 +5,14 @@
 
 ## Version
 
-0.2.2
+0.2.3
 
 ### Changelog
+
+- 0.2.3
+
+  - Fix stated behavior if doing nothing
+    as defined in [how to fail](../changes/2020-06-04_how-to-fail-with-keyrings/change.md).
 
 - 0.2.2
 
@@ -106,8 +111,8 @@ and MAY modify it with any of the following behaviors:
 If this keyring attempted any of the above behaviors, and successfully completed those behaviors,
 it MUST output the modified [encryption materials](structures.md#encryption-materials).
 
-If the keyring did not attempt any of the above behaviors, it MUST output the
-[encryption materials](structures.md#encryption-materials) unmodified.
+If the keyring did not attempt any of the above behaviors, it MUST fail
+and it MUST NOT modify the [encryption materials](structures.md#encryption-materials).
 
 #### Generate Data Key
 
@@ -159,7 +164,8 @@ MAY modify it with the following behavior:
 If this keyring attempted the above behavior, and succeeded, it MUST output the modified [decryption materials](structures.md#decryption-materials).
 
 If the keyring did not attempt the above behavior,
-the keyring MUST output the [decryption materials](structures.md#decryption-materials) unmodified.
+the keyring MUST fail
+and MUST NOT modify the [decryption materials](structures.md#decryption-materials).
 
 #### Decrypt Data Key
 
--- a/framework/keyring-interface.md
+++ b/framework/keyring-interface.md
@@ -113,8 +113,6 @@ it MUST output the modified [encryption materials](structures.md#encryption-mate
 
 If the keyring did not attempt any of the above behaviors, it MUST fail
 and it MUST NOT modify the [encryption materials](structures.md#encryption-materials).
-In failing, it MUST provide the failure message
-`No action taken by keyring on encrypt.`
 
 #### Generate Data Key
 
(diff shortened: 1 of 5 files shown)


framework/keyring-interface.md
```diff
@@ -106,8 +111,10 @@ and MAY modify it with any of the following behaviors:
 If this keyring attempted any of the above behaviors, and successfully completed those behaviors,
 it MUST output the modified [encryption materials](structures.md#encryption-materials).
 
-If the keyring did not attempt any of the above behaviors, it MUST output the
-[encryption materials](structures.md#encryption-materials) unmodified.
+If the keyring did not attempt any of the above behaviors, it MUST fail
+and it MUST NOT modify the [encryption materials](structures.md#encryption-materials).
+In failing, it MUST provide the failure message
+`No action taken by keyring on encrypt.`
```
lavaleri, line 117: This feels like a SHOULD, or maybe soften the requirement in some other way. Specifically this feels not easy to strictly implement for C.

Also, this isn't added in the specific raw keyring specs, so do we actually care about this requirement?

mattsb42-aws: > This feels like a SHOULD, or maybe soften the requirement in some other way. Specifically this feels not easy to strictly implement for C.

C has a mechanism for communicating a message when something fails. I intentionally avoided "error message"/etc; would different wording make more sense?

mattsb42-aws: > Also, this isn't added in the specific raw keyring specs, so do we actually care about this requirement?

My intent was to have this defined in a single place; if you think that each keyring doc needs to reinforce this requirement, I can add that.

seebees: Perhaps more about what the intention is?  The goal here is to have a single universally identifiable way to see "nothing happened".  Maybe the MUST is that customers can identify this case, and the SHOULD is the specific error message?  Or maybe "provide" is the tricky part? "Able to find" might be a reasonable description?

lavaleri: ^ I think I like focusing on making sure the customer gets the same message. Maybe this is really nitpicky, but in C's case it's not really the Keyring providing this message, but whatever is using the keyring being able to understand this sort of error and then make sure that the correct message gets to customers. Maybe one of our C people can just take a look and make sure that we're expressing the right intention here.

lavaleri: > My intent was to have this defined in a single place; if you think that each keyring doc needs to reinforce this requirement, I can add that.

I don't like duplication, but in the past people implementing specific keyrings have been confused by requirements that exist in the interface but not the specific keyring, so I think we should add it, otherwise it's going to get missed.

robin-aws: On further reflection, I don't think specifying this exact condition/error message in the specification is useful. 

Now that we've eliminated the no-op result as an option and just have success or failure, I don't think it's meaningful to talk about whether a keyring "attempted" the behaviour or not. In every case I can think of (e.g. a KMS keyring not attempting decryption because of a region mis-match) you should be able to say something more specific than just "nothing happened" (e.g. "CMK is not in us-east-1") which will be much more helpful for debugging issues. Otherwise a multi-keyring of KMS keyrings is just going to produce N "nothing happened" messages, and that's not going to be enough to be helpful.

I can understand the intent here is that keyrings need to communicate an outcome that might have been a no-op before SOMEHOW, but the earlier change already addresses that by saying they are now just failures, and each implementation will have a way to diagnose/debug failures. 

I DO think there's a good tenet buried in here along the lines of "provide as much useful debugging information as you can" - is this something we can write down somewhere else?

robin-aws: > ^ I think I like focusing on making sure the customer gets the same message. Maybe this is really nitpicky, but in C's case it's not really the Keyring providing this message, but whatever is using the keyring being able to understand this sort of error and then make sure that the correct message gets to customers. Maybe one of our C people can just take a look and make sure that we're expressing the right intention here.

To address this question specifically: "produce" is vague enough to still apply, since a keyring can emit a log message in an equivalent way. For "expected" failures you probably want them to be sufficiently quiet (i.e. "info" level). To have as much debugging information as a tree of error values you'd need extra identifiers or qualifying information so you know where exactly they are coming from, but that ends up being safely in the implementation-specific domain IMO (since this is a generic C problem, or perhaps aws-c-common).

mattsb42-aws: After reading through this thread and talking with @seebees offline, I'm just going to pull the error message statements from this change. I still think standardized error messages SHOULD be defined in the future, but I am now convinced that defining these properly and well for each keyring is outside of the scope of this change.

PR #225, opened 2020-08-11
chore: clarify wording for CMM algorithm suites

alex-chew:
_Issue #, if available:_ N/A

_Description of changes:_
In the CMM interface, the algorithm suite is optional in the Encryption Materials request and mandatory in the Decrypt Materials request. This PR clarifies that specification statements about the algorithm suite apply only if the algorithm suite is present (in the Encryption Materials case), and that such statements are unconditional when the algorithm suite is mandatory (i.e. in the Decrypt Materials case).

### DIFF
--- a/framework/caching-cmm.md
+++ b/framework/caching-cmm.md
@@ -135,7 +135,8 @@ The number of bytes encrypted by the [encryption](structures.md#encryption-mater
 
 ### Get Encryption Materials
 
-If the [algorithm suite](algorithm-suites.md) requested contains a [Identity KDF](algorithm-suites.md#identity-kdf),
+If the [algorithm suite](algorithm-suites.md) is present in the encryption materials request,
+and it contains an [Identity KDF](algorithm-suites.md#identity-kdf),
 the caching CMM MUST obtain the encryption materials by making a call to the underlying CMM's [Get Encryption Materials](cmm-interface.md#get-encryption-materials) function.
 
 Otherwise, the caching CMM MUST attempt to find the [encryption materials](structures.md#encryption-materials)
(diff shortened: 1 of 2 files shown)


framework/cmm-interface.md
```diff
@@ -106,17 +106,19 @@ The encryption materials returned MUST include the following:
 - [Encryption Context](structures.md#encryption-context)
   - The CMM MAY modify the encryption context.
 
-If the algorithm suite contains a [signing algorithm](algorithm-suites.md#signature-algorithm):
-
-- The CMM MUST include a [signing key](structures.md#signing-key).
```
lavaleri, line 111: This is referring to the returned algorithm suite, which is required.

framework/caching-cmm.md
```diff
@@ -135,7 +135,8 @@ The number of bytes encrypted by the [encryption](structures.md#encryption-mater
 
 ### Get Encryption Materials
 
-If the [algorithm suite](algorithm-suites.md) requested contains a [Identity KDF](algorithm-suites.md#identity-kdf),
+If the [algorithm suite](algorithm-suites.md) is present in the encryption materials request,
+and it contains an [Identity KDF](algorithm-suites.md#identity-kdf),
```
lavaleri, line 139: What happens if the algorithm suite isn't present? What does the Caching CMM do?

framework/cmm-interface.md
alex-chew: I see, I'll clarify that.

framework/caching-cmm.md
alex-chew: The next paragraph ("Otherwise, the caching CMM...") addresses that: If either the algorithm suite isn't present, or if it is present and doesn't contain an Identity KDF, then the caching CMM checks the cache, and then returns materials if found or delegates to the underlying CMM if not.

framework/caching-cmm.md
```diff
@@ -147,10 +148,12 @@ If a cache entry is found, the caching CMM MUST return the encryption materials
 If a cache entry is not found or the cache entry is expired, the caching CMM MUST then attempt to obtain the encryption materials
 by making a call to the underlying CMM's [Get Encryption Materials](cmm-interface.md#get-encryption-materials).
 
-If the [algorithm suite](algorithm-suites.md) requested does not contain an [Identity KDF](algorithm-suites.md#identity-kdf),
+If the [algorithm suite](algorithm-suites.md) is present in the encryption materials request,
+and it does not contain an [Identity KDF](algorithm-suites.md#identity-kdf),
 the caching CMM MUST add the encryption materials obtained from the underlying CMM into the underlying CMC.
```
lavaleri, line 153: This is missing the "If no algorithm suite is present" case, in which we still want to add the materials to the underlying CMM (using the no-alg identifier explained below).

alex-chew: Good catch. Does this look good?

```suggestion
If the [algorithm suite](algorithm-suites.md) is present in the encryption materials request
and does not contain an [Identity KDF](algorithm-suites.md#identity-kdf),
or if the algorithm suite is not present,
then the caching CMM MUST add the encryption materials obtained from the underlying CMM into the underlying CMC.
```

lavaleri: LGTM. thanks!

PR #234, opened 2021-06-17
fix: Java linter interaction with test_conditions

seebees:
If the linter moves the test `DisplayName`
to a new line because of the line length,
the condition should not break.

### DIFF
--- a/util/test_conditions
+++ b/util/test_conditions
@@ -31,7 +31,7 @@ const fileSettings = {
     complianceConfig: standardComplianceParams,
   },
   ".java:": {
-    conditionRemove: /(\*\/)|(\/\*)|(@DisplayName\(\")|(\"\))/g,
+    conditionRemove: /(\*\/)|(\/\*)|(@DisplayName\(\")|(\"\))|(\")|(\"\))/g,
     complianceConfig: standardComplianceParams,
   },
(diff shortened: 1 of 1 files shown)


util/test_conditions
```diff
@@ -31,7 +31,7 @@ const fileSettings = {
     complianceConfig: standardComplianceParams,
   },
   ".java:": {
-    conditionRemove: /(\*\/)|(\/\*)|(@DisplayName\(\")|(\"\))/g,
+    conditionRemove: /(\*\/)|(\/\*)|(@DisplayName\(\")|(\"\))|(\")|(\"\))/g,
```
lavaleri, line 34: `(\"\))` is included twice

seebees: It is `"` and `")`
Do you have a suggestion on a better RegEx? :)
It is trying to match

`      "Precondition: Something here.")`
and turn that into
`Precondition: Something here.`

lavaleri: I mean I think it should be
```suggestion
    conditionRemove: /(\*\/)|(\/\*)|(@DisplayName\(\")|(\"\))|(\")/g,
```
`(\"\))` is already in the regex once

lavaleri (approved):

PR #239, opened 2022-05-11
Adjust link to point to `algorithm-suites` instead of `algorithm-suite`

pgermishuys:
_Issue #, if available:_

_Description of changes:_
The current link is pointing to a location that doesn't exist (resulting in a 404). The commit fixes the link to a known location for the algorithm suites.


texastony (approved):
@pgermishuys 

Thank you for this correction!
This PR looks good to me.

PR #254, opened 2023-06-14
chore: update CMC for multi-threading

ajewellamz:
_Issue #, if available:_

_Description of changes:_

# Check any applicable:

### DIFF
--- a/framework/cryptographic-materials-cache.md
+++ b/framework/cryptographic-materials-cache.md
@@ -147,10 +147,10 @@ and an unsuccessful call returns a cache miss.
 
 If used in a multi-threaded context :
 
- * Get Cache Entry MAY return a cache miss when the TTL has net yet been exceeded.
+- Get Cache Entry MAY return a cache miss when the TTL has net yet been exceeded.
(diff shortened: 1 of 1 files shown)


texastony (commented):
Question: Should there not be some general statement like:
`The Cryptographic Materials Cache MUST support Multiple Threads.`

ajewellamz:
> Question: Should there not be some general statement like: `The Cryptographic Materials Cache MUST support Multiple Threads.`

I'm not sure. CMC is an abstract class, Maybe we want to allow an implementation that is optimized for single threaded use. The plan is already to have `CreateCryptographicMaterialsCache` return either a `SynchronizedLocalCMC` or a `StormTrackingCMC`. Maybe returning a `LocalCMC` would be appropriate under some circumstances.

I don't know the right answer.

texastony:
Hmm... Maybe we should have a more generic requirement.
i.e:
`The MPL MUST include a CMC implementation that supports Multiple Threads.`

texastony (approved):

PR #256, opened 2023-06-20
chore: update keyring interface to reflect KMS keyring changes

justplaz:
_Issue #, if available:_

_Description of changes:_

Minimal updates to the `keyring-interface` specification to reflect that there are now multiple AWS KMS keyrings.  

# Check any applicable:

### DIFF
--- a/framework/keyring-interface.md
+++ b/framework/keyring-interface.md
@@ -215,7 +215,7 @@ the decryption is overwhelmingly likely to fail.
 Users SHOULD use a keyring that protects wrapping keys and performs cryptographic operations within a secure boundary.
 Examples are:
 
-- The built-in [AWS KMS keyrings](./aws-kms),
+- The built-in [AWS KMS keyrings](./aws-kms/),
   which use AWS Key Management Service (AWS KMS) customer master keys (CMKs) that never leave AWS KMS as plaintext.
 - A custom keyring that uses wrapping keys that are stored in your hardware security modules (HSMs)
 - A custom keyring protected by another master key service.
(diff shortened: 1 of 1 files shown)


framework/keyring-interface.md
```diff
@@ -93,7 +97,7 @@ and SHOULD be equal to a UTF-8 encoding of the key name.
 
 ## Supported Keyrings
 
-- [AWS KMS Keyring](kms-keyring.md)
+- [AWS KMS Keyrings](./aws-kms/)
```
seebees, line 100: I think that this is "fine".
The idea was to link to the "real" keyring.
There are like 4 or so now...
This link will go to "all the KMS stuff" as opposed to the specific keyrings...

justplaz: Indeed, this is the smallest change that fixes the links to point to a "more" correct location. 
If you're keen, I can make a larger change to organize to something like:

```
framework/keyrings/keyring-interface.md
framework/keyrings/multi-keyring.md
...
framework/keyrings/aws-kms/aws-kms-keyring.md
framework/keyrings/aws-kms/aws-kms-mrk-discovery-keyring.md
...
```

seebees: Let's make a full list.
At this point there are quite a few
and I can see us missing one...
It would be good to have a nice single list _somewhere_.

I don't think we need to keep this list _everywhere_
So maybe we just update the list here, and leave the rest of the links the same

framework/keyring-interface.md
```diff
@@ -79,7 +83,7 @@ The key provider ID MUST be a binary value
 and SHOULD be equal to a UTF-8 encoding of the key namespace.
 
 This value MUST NOT be "aws-kms"
```
texastony, line 85: Issue/Suggestion: The new MPL introduces the
["aws-kms-hierarchy"](https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/Constants.dfy#L18-L23) & ["aws-kms-rsa"](https://github.com/aws/aws-cryptographic-material-providers-library-java/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/Constants.dfy#L25C21-L29) key provider IDs.

```suggestion
This value MUST NOT be or start with "aws-kms"
```

texastony (approved):
I request one change. 
But in general, I approve.

texastony (approved):

framework/keyring-interface.md
```diff
@@ -211,8 +223,8 @@ the decryption is overwhelmingly likely to fail.
 Users SHOULD use a keyring that protects wrapping keys and performs cryptographic operations within a secure boundary.
 Examples are:
 
-- The built-in [AWS KMS keyring](kms-keyring.md),
-  which uses AWS Key Management Service (AWS KMS) customer master keys (CMKs) that never leave AWS KMS plaintext.
+- The built-in [AWS KMS keyrings](./aws-kms/),
+  which use AWS Key Management Service (AWS KMS) customer master keys (CMKs) that never leave AWS KMS as plaintext.
```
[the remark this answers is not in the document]
justplaz: It's not 100% accurate, but also not necessarily incorrect. 
The authoritative reference for KMS key types is here: https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html

Originally, CMK (Customer Master Key) referred to the "top level" key, 
i.e. the thing you create when you call the `CreateKey` API,
distinct from say a data key or an HBK. 
The set of CMKs comprised customer managed keys,
as well as AWS managed keys and AWS owned keys. 
A few years ago, 
AWS KMS renamed CMKs to AWS KMS Keys, 
and remapped the CMK acronym to Customer Managed Keys, 
effectively partitioning the set of KMS Keys. 
However, no one outside of AWS can create or manage an AWS Managed Key,
so for all public intents and purposes, 
particularly relating to the AWS Encryption SDK,
they are the same thing. 

All of this to say,
we should update our spec (and all documentation) 
to reflect the changes described above.
However, that work is not in scope for this PR,
and must be prioritized accordingly.

PR #257, opened 2023-06-23
fix: Minor fixes to Required Encryption Context CMM

lavaleri:
Clarify wording in `changes` in order to not imply that we are asserting that data that is sent along as AAD is protected. The importance to focus on is that customers may have data that is not sensitive, and is ok to process as AAD, but they would still prefer to not have that data stored alongside the message in readily available plaintext.

Also fix the naming in the Required Encryption Context CMM.

# Check any applicable:


seebees (approved):
LGTM

PR #261, opened 2023-09-27
chore(decrypt): specify behavior on decrypt with encryption context

josecorella:
_Issue #, if available:_

_Description of changes:_

# Check any applicable:

### DIFF
--- a/changes/2023-09-26_encryption-context-decrypt/background.md
+++ b/changes/2023-09-26_encryption-context-decrypt/background.md
@@ -1,0 +1,540 @@
+[//]: # "Copyright Amazon.com Inc. or its affiliates. All Rights Reserved."
+[//]: # "SPDX-License-Identifier: CC-BY-SA-4.0"
+
+# Which Encryption Context should be used in Decrypt Operations
+
+## Definitions
+
+### Conventions used in this document
+
+The key words
+"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Issues and Alternatives
+
+### What kind of Encryption Context should Decrypt return?
+
+Decrypt returns Encryption Context.
+This requirement may be satisfied by returning the parsed header that contains this value.
+
+With the addition of the [required encryption context cmm](../../framework/required-encryption-context-cmm.md),
+the encrypt API is able to filter out encryption context key-value pairs that
+are not stored but instead authenticated.
+
+Currently, on decrypt takes in optional encryption context.
+This optional encryption context is used as encryption context for authentication only.
+
+The Required Encryption Context CMM can modify the encryption context on decrypt.
+This results in values that are not authenticated against the message or the plaintext data key.
+
+It MUST be required that the decryption materials obtained from the underlying CMM's MUST
+contain all configured encryption context keys in its encryption context and the values MUST remain
+unchanged.
+
+The Decrypt API MUST return the encryption context stored in the header and the encryption
+context used for authentication only.
+
+### What changes to the existing Required Encryption Context CMM are required?
+
+None
+
+### API changes
+
+If required encryption context keys match any keys stored in the message header,
+their values MUST match. Otherwise old messages are not compatible with the required encryption context CMM.
+
+```
+forall k <- decryptionMaterials.requiredEncryptionContextKeys
+   | k in headerEncryptionContext
+   :: decryptionMaterials[k] == headerEncryptionContext[k]
+```
+
+## Testing
+
+In order to test that decrypt behaves as expected according to the specification, testing should include the following
+scenarios to accurately reason about the behavior of using the Required Encryption Context CMM.
+
+### Test Decrypt with Identical Encryption Context
+
+Without using a Required Encryption Context CMM, Decrypt MUST succeed if provided with the identical encryption context
+that was used on Encrypt.
+
+```
+// The string "asdf" as bytes
+var asdf := [ 97, 115, 100, 102 ];
+var aesKeyring := GetAesKeyring();
+
+// Test supply same encryption context on encrypt and decrypt NO filtering
+var encryptionContext := map[ keyA := valA, keyB := valB ];
+
+var encryptOutput := esdk.Encrypt(Types.EncryptInput(
+   plaintext := asdf,
+   encryptionContext := Some(encryptionContext),
+   materialsManager := None,
+   keyring := Some(aesKeyring),
+   algorithmSuiteId := None,
+   frameLength := None
+));
+
+expect encryptOutput.Success?;
+var esdkCiphertext := encryptOutput.value.ciphertext;
+
+var decryptOutput := esdk.Decrypt(Types.DecryptInput(
+   ciphertext := esdkCiphertext,
+   materialsManager := None,
+   keyring := Some(aesKeyring),
+   encryptionContext := Some(encryptionContext)
+));
+
+expect decryptOutput.Success?;
+var cycledPlaintext := decryptOutput.value.plaintext;
+expect cycledPlaintext == asdf;
+```
+
+### Test Filter out Encryption Context on Encrypt AND supply on Decrypt
+
+On Encrypt will not write one key-value pair to the message header and instead will include it in the header signature.
+On Decrypt will supply the Encryption Context pair that was not written but will
+NOT use the Required Encryption Context CMM. This operation MUST succeed.
+
+```
+// The string "asdf" as bytes
+var asdf := [ 97, 115, 100, 102 ];
+var aesKeyring := GetAesKeyring();
+
+// Test supply same encryption context on encrypt and decrypt NO filtering
+var encryptionContext := map[ keyA := valA, keyB := valB ];
+var reproducedEncryptionContext := map[keyA := valA];
+var requiredEncryptionContextKeys := [keyA];
+
+var defaultCMM :- expect mpl.CreateDefaultCryptographicMaterialsManager(
+   mplTypes.CreateDefaultCryptographicMaterialsManagerInput(
+         keyring := aesKeyring
+   )
+);
+
+// Create Required EC CMM with the required EC Keys we want
+var reqCMM :- expect mpl.CreateRequiredEncryptionContextCMM(
+   mplTypes.CreateRequiredEncryptionContextCMMInput(
+         underlyingCMM := Some(defaultCMM),
+         keyring := None,
+         requiredEncryptionContextKeys := requiredEncryptionContextKeys
+   )
+);
+
+var encryptOutput := esdk.Encrypt(Types.EncryptInput(
+   plaintext := asdf,
+   encryptionContext := Some(encryptionContext),
+   materialsManager := Some(reqCMM),
+   keyring := None,
+   algorithmSuiteId := None,
+   frameLength := None
+));
+
+expect encryptOutput.Success?;
+var esdkCiphertext := encryptOutput.value.ciphertext;
+
+var decryptOutput := esdk.Decrypt(Types.DecryptInput(
+   ciphertext := esdkCiphertext,
+   materialsManager := None,
+   keyring := Some(rawAesKeyring),
+   encryptionContext := Some(reproducedEncryptionContext)
+));
+
+expect decryptOutput.Success?;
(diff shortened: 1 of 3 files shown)


changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,540 @@
...
+
+expect decryptOutput.Success?;
+var cycledPlaintext := decryptOutput.value.plaintext;
+expect cycledPlaintext == asdf;
+```
+
+### Test Different Encryption Context on Decrypt Failure
+
+Encrypt with and store all encryption context in the message header.
+On Decrypt will supply additional encryption context not stored in the header.
```
lavaleri, line 281: This is testing a total disjoint between expected EC and inputted EC. We should seperately test "inputted EC is missing a pair", "inputted EC has an incorrect value for one pair", and "inputted EC has an extra pair".

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,540 @@
...
+
+```
+// The string "asdf" as bytes
+var asdf := [ 97, 115, 100, 102 ];
+var aesKeyring := GetAesKeyring();
+
+// Test supply same encryption context on encrypt and decrypt NO filtering
+var encryptionContext := map[ keyA := valA, keyB := valB ];
+var reproducedAdditionalEncryptionContext := map[ keyA := valB, keyB := valA  ];
+var requiredEncryptionContextKeys := [keyA, keyB];
```
lavaleri, line 345: I thought this was testing context that _is_ in the header?

changes/2023-09-26_encryption-context-decrypt/background.md
lavaleri: Looking at the tests below, I think we cover all of these except "inputted EC has an extra pair"

changes/2023-09-26_encryption-context-decrypt/change.md
```diff
@@ -0,0 +1,52 @@
...
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Summary
+
+The [Decrypt](../../client-apis/decrypt.md) API does not specify
+how it should deal with required encryption context keys it receives from the call to the
+Cryptographic Materials Manager, the encryption context it optionally takes in, and the stored
+encryption context on the message header.
+
+Decrypt MUST verify that if the keys in the required encryption context keys exist in the reproduced encryption context
```
lavaleri, line 36: > if the keys in the required encryption context keys exist in the reproduced encryption context

is this not true already?

changes/2023-09-26_encryption-context-decrypt/change.md
```diff
@@ -0,0 +1,52 @@
...
+that if a [required encryption context key](../../framework/structures.md#required-encryption-context-keys)
+exists in the message [header's AAD](../../data-format/message-header.md#aad)
+the value in the [decryption materials' encryption context](../../framework/structures.md#decryption-materials)
+MUST match what is store in the message header.
+
+This verifies that the reproduced encryption context is used to decrypt the message and
+NOT some key-value pair that the underlying CMM MAY have modified.
+
+If required keys exist, match stored encryption context, AND decryption succeeds
+Decrypt MUST return both the Encryption Context stored in the header AND the
```
lavaleri, line 51: This "AND" is ambiguous. I believe it should be the union of these maps.

client-apis/decrypt.md
```diff
@@ -161,7 +163,8 @@ that implementation SHOULD NOT provide an API that allows the caller to stream t
 ### Encryption Context
 
 The [encryption context](../framework/structures.md#encryption-context) that is used as
-additional authenticated data during the decryption of the input [encrypted message](#encrypted-message).
+additional authenticated data during the decryption of the input [encrypted message](#encrypted-message) and
+the [encryption context for authentication only](../client-apis/encrypt.md#encryption-context-not-stored-in-the-message) if available.
```
lavaleri, line 167: It's all used as AAD during the decryption, so just appending an "and ..." here doesn't make sense.
```suggestion
additional authenticated data during the decryption of the input [encrypted message](#encrypted-message).
Specifically, the union of the encryption context serialized into the message header and
the encryption context for authentication only, if available.
```

lavaleri: Sprinkle a MUST in there as well, so it is easier to duvet.

lavaleri (changes requested):

changes/2023-09-26_encryption-context-decrypt/change.md
lavaleri: Actually, should it just replicate our logic for inclusion in the header validation?
i.e. Every key-value pair in the header, and every "required" key-value pair in the EC?

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,540 @@
...
+
+Currently, on decrypt takes in optional encryption context.
+This optional encryption context is used as encryption context for authentication only.
+
+The Required Encryption Context CMM can modify the encryption context on decrypt.
+This results in values that are not authenticated against the message or the plaintext data key.
+
+It MUST be required that the decryption materials obtained from the underlying CMM's MUST
+contain all configured encryption context keys in its encryption context and the values MUST remain
+unchanged.
```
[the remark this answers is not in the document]
josecorella: "configured encryption conext key" this should be "required encryption context key"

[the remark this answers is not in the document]
josecorella: I am trying to get at is that if the CMM returns encryption context which IS stored AND one of the keys in the required encryption context is also in this encryption context then the value in their key-value pair mapping MUST remain unchanged

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,540 @@
...
+context used for authentication only.
+
+### What changes to the existing Required Encryption Context CMM are required?
+
+None
+
+### API changes
+
+If required encryption context keys match any keys stored in the message header,
+their values MUST match. Otherwise old messages are not compatible with the required encryption context CMM.
```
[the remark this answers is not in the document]
josecorella: I may remove the second sentence. I am trying to say that without the check you will not be able to read messages that were written without a required encryption context cmm

changes/2023-09-26_encryption-context-decrypt/background.md
josecorella: adding in latest revision

changes/2023-09-26_encryption-context-decrypt/background.md
josecorella: yup good catch

changes/2023-09-26_encryption-context-decrypt/change.md
josecorella: It is not called out explicitly so we are just trying to close the ambiguity

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,584 @@
...
+If any required encryption context keys from the CMM's `DecryptionMaterials` response
+match any keys stored in the message header, their values MUST match.
+
+```
+forall k <- decryptionMaterials.requiredEncryptionContextKeys
+   | k in headerEncryptionContext
+   :: decryptionMaterials[k] == headerEncryptionContext[k]
+```
+
+## Testing
```
lavaleri, line 55: Thinking more about this, I think it would be better to more clearly define what the matrix is we care about, as opposed to listing out the different tests and code for them. I would prefer we cut the example code, but make it extremely clear what every case we need to check is in a way that makes it easy to reason that it is complete. For example, maybe frame it like this:

---

First, suppose we have a message where {A:B,C:D} is in the header, and {E:F} is not stored, but is included in the authentication.

For this message, go through the following test cases:
- customer supplies correct required key-value {E:F} (SUCCESS)
- customer supplies correct required key-value and additional correct header value {A:B,E:F} (SUCCESS)
- customer supplies does not supply required key-value {} (FAILURE)
- customer supplies incorrect value for required key-value {E:X} (FAILURE)
- customer supplies correct required key-value and additional incorrect header value {A:X,E:F} (FAILURE)
- customer supplies correct required key-value and additional incorrect, non-header key-value {E:F,X:Y} (FAILURE)

For every one of the above cases, attempt to decrypt using both a Default CMM, and a REC CMM with E configured as a "required key".

Next, suppose we have a message where {A:B,C:D,E:F} is in the header, and there is no additional non-stored key-values authenticated with the message.

For this message, go through the following test cases:
- customer supplies no reproduced context {} (SUCCESS)
- customer supplies additional correct header values {A:B,C:D,E:F} (SUCCESS)
- customer supplies incorrect value for key in header {A:X} (FAILURE)
- customer supplies incorrect, non-header key-value {X:Y} (FAILURE)

For each of these use a Default CMM.

---

This is more test cases than we currently have, and we may be able to convince ourselves that some of those test cases are not actually useful. However, I would prefer if this document focused on listing the parameters for the test cases in a way that is easy for readers to see whether we are missing any (And if we decide any are not useful, say why). If we do this, I don't think we need to give example code for each test case.

client-apis/decrypt.md
```diff
@@ -162,6 +164,8 @@ that implementation SHOULD NOT provide an API that allows the caller to stream t
 
 The [encryption context](../framework/structures.md#encryption-context) that is used as
 additional authenticated data during the decryption of the input [encrypted message](#encrypted-message).
+Specifically, it MUST be the union of the encryption context serialized into the message header and
+the encryption context for authentication only, if available.
```
lavaleri, line 168: `encryption context for authentication only` Can we point to the definition of this somewhere?

lavaleri (changes requested):

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,584 @@
...
+
+In order to test that decrypt behaves as expected according to the specification, testing should include the following
+scenarios to accurately reason about the behavior of using the Required Encryption Context CMM.
+
```
seebees, line 59: I suggest a truth table here.
I'm working on how this would look ATM.

changes/2023-09-26_encryption-context-decrypt/change.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Summary
+
+The [Decrypt](../../client-apis/decrypt.md) API does not specify
+how it should deal with required encryption context keys it receives from the call to the
+Cryptographic Materials Manager, the encryption context it optionally takes in, and the stored
+encryption context on the message header.
+
+Decrypt MUST verify that if the keys in the required encryption context keys exist in the reproduced encryption context
+AND they are stored in the message header, they MUST have the same value as the reproduced encryption context supplied.
```
seebees, line 37: This is not how you have change the spec:

```
- If a [required encryption context key](../framework/structures.md#required-encryption-context-keys) in
  the decryption materials' response exists in the [message header body](../data-format/message-header.md#header-body)
  the value in the decryption materials' encryption context MUST equal the value stored in the message header body.
```

If you want what you have specified, then I think you need:

```suggestion
For required encryption context keys the value in stored in the message header
MUST equal the value in decryption materials encryption context.
```

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -17,568 +17,139 @@ in this document are to be interpreted as described in
...
+- `{ a: c }`
+- `{ b: c }`
+- `{ a: c, b: b }`
+- `{ a: c, b: c }`
+- `{ c: c }`
+- `{ a: a, c: c }`
+- `{ b: b, c: c}`
+- `{ a: a, b: b, c: c }`
+
+### Message: `{a: a, b: b}` / `{}`
```
seebees, line 106: I think we can further wordsmith these tables to simplify.
But this is a good start!

changes/2023-09-26_encryption-context-decrypt/background.md
```diff
@@ -0,0 +1,155 @@
...
+Given the EC `{ a: a, b: b }`
+we break this into the following interesting options:
+
+Stored EC/Not Stored EC
+
+- `{a: a, b: b}` / `{}`
+- `{a: a}` / `{b: b}`
+- `{}` / `{a: a, b: b}`
+
+CMM Material/Required Keys
```
lavaleri, line 83: This is missing:
- `{a: a, b: b, x: x}` / `{a}`
- `{a: x, b: b}` / `{a}`

PR #277, opened 2024-08-08
chore(hkeyring): Update HKeyring cache identifier formula in-line with dafny implementation

RitvikKapila:
_Issue #, if available:_
NA

_Description of changes:_
The dafny implementation of cache identifier formulae of the HKeyring aren't exactly coherent with the spec. This PR corrects that by changing the spec in accordance with the dafny code of the HKeyring.

# Check any applicable:

### DIFF
--- a/framework/aws-kms/aws-kms-hierarchical-keyring.md
+++ b/framework/aws-kms/aws-kms-hierarchical-keyring.md
@@ -293,18 +293,18 @@ When the hierarchical keyring receives an OnEncrypt request,
 the cache entry identifier MUST be calculated as the first 32 bytes of the
 SHA-512 hash of the following byte strings, in the order listed:
 
-| Field                    | Length (bytes) | Interpreted as |
-| ------------------------ | -------------- | -------------- |
-| Length of branch-key-id  | 3              | UInt8          |
-| branch-key-id            | Variable       | UTF-8 Encoded  |
-| Null Byte                | 1              | `0x00`         |
-| Constant string "ACTIVE" | 6              | UTF-8 Encoded  |
+| Field                    | Length (bytes) | Interpreted as         |
+| ------------------------ | -------------- | -----------------------|
+| Length of branch-key-id  | 3              | UInt8                  |
+| SHA-512(branch-key-id)   | 64             | SHA-512(UTF-8 Encoded) |
+| Null Byte                | 1              | `0x00`                 |
+| Constant string "ACTIVE" | 6              | UTF-8 Encoded          |
 
 As a formula:
 
 ```
 branch-key-id = UTF8Encode(hierarchicalKeyring.BranchKeyIdentifier)
-branch-key-digest - SHA512(branch-key-id)
+branch-key-digest = SHA512(branch-key-id)
 
 ENTRY_ID = SHA512(
     LengthUint8(branch-key-id) +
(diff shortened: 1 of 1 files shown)


framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -328,13 +328,12 @@ it MUST calculate the cache entry identifier as the first 32 bytes of the SHA-51
...
 branch-key-id = UTF8Encode(edk.providerInfo)
-branch-key-digest - SHA512(branch-key-id)
 
 ENTRY_ID = SHA512(
     LengthUint8(branch-key-id) +
-    branch-key-digest +
+    branch-key-id +
     0x00 +
-    branch key version
+    UTF8Encode(branch-key-version)
```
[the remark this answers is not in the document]
RitvikKapila: Yes, good catch! Thanks.

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -293,18 +293,18 @@ When the hierarchical keyring receives an OnEncrypt request,
...
-| Field                    | Length (bytes) | Interpreted as |
-| ------------------------ | -------------- | -------------- |
-| Length of branch-key-id  | 3              | UInt8          |
-| branch-key-id            | Variable       | UTF-8 Encoded  |
-| Null Byte                | 1              | `0x00`         |
-| Constant string "ACTIVE" | 6              | UTF-8 Encoded  |
+| Field                    | Length (bytes) | Interpreted as         |
+| ------------------------ | -------------- | ---------------------- |
+| Length of branch-key-id  | 3              | UInt8                  |
+| SHA-512(branch-key-id)   | 64             | SHA-512(UTF-8 Encoded) |
```
texastony, line 299:
```suggestion
| SHA-512(branch-key-id)   | 64             | bytes |
```

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -293,18 +293,18 @@ When the hierarchical keyring receives an OnEncrypt request,
 the cache entry identifier MUST be calculated as the first 32 bytes of the
 SHA-512 hash of the following byte strings, in the order listed:
```
josecorella, line 294: imo; it would be best to split these up similar to how we write the requirements of other structures or orders throughout the spec since it can be difficult to add duvet annotations to these charts.
For example,
```suggestion
the cache entry identifier MUST be calculated as the first 32 bytes of the
SHA-512 hash of the following byte strings, in the order listed:
- MUST be the length of the branch key id in its uint8 interpreatation
- MUST be the SHA512 output of the UTF8 encoded branch-key-id
- MUST be a single null byte `0x00`
- MUST be the constant UTF8 encoded string "ACTIVE" 
```

This way we can easily duvet the implementation.

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Yeah I was wondering how we should write this. This is helpful. Thanks!

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Sounds good. Thanks for pointing this out.

josecorella (approved):
lgtm

texastony (approved):

PR #279, opened 2024-08-14
feat(Caching): Cache across Hierarchy Keyring

RitvikKapila:
_Issue #, if available:_

_Description of changes:_

# Check any applicable:


changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -43,49 +51,41 @@ By "KMS Relationship", we mean any or all of the following:
...
-## Summary
+A [Cryptographic Materials Cache](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/cryptographic-materials-cache.md) can be used by a Hierarchy Keyring, a Caching CMM, and the DB-ESDK.
+It can store four types of [materials](https://github.com/aws/aws-cryptographic-material-providers-library/blob/c07a51fc29ff70411f7573bca96d2a091db8c1ed/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/cryptographic-materials-cache.smithy#L89) for three different use-cases:
 
-To improve the MPL Consumer data key caching expierence,
-when using the Hierarchical Keyring,
-we need to allow caching across KeyStores/KMS Clients/KMS Keys.
+- Hierarchy Keyring: keyStore#BranchKeyMaterials
+- CachingCMM: EncryptionMaterials and DecryptionMaterials
+- DB-ESDK: keyStore#BeaconKeyMaterials
```
texastony, line 61: Style: `#` is an odd joiner in Markdown. Are you trying to create a link?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -43,49 +51,41 @@ By "KMS Relationship", we mean any or all of the following:
...
-the CMC MAY cache Branch Keys protected by different
-KMS Relationships.
+- Only one Hierarchy Keyring can use one cache,
+  i.e., currently, we do NOT support a shared cache
+  between multiple Hierarchy Keyrings.
+- Multiple CachingCMMs can use a single shared cache
+  (this is for native languages, there is currently no
+  CachingCMM implemented in dafny)
+- We do NOT support a shared cache between multiple
+  DB-ESDK clients.
```
texastony, line 74: Issue/Suggestion/Blocking: The DB-ESDK client can, theoretically, consume CachingCMMs and Hierarchy Keyrings.

Where you write DB-ESDK, can you replace it with Searchable Encryption?

```suggestion
- We do NOT support a shared cache between multiple
  Searchable Encryption Configurations.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -144,7 +145,210 @@ is to make the Hierarchy Keyring
...
+we MUST break the Cryptographic Materials Cache (CMC)
+out of the Hierarchy Keyring.
+
+By allowing MPL Consumers to provide an already initialized CMC
+to the Hierarchy Keyring at construction,
+the CMC MAY cache Branch Keys protected by different
+KMS Relationships.
+
+This simplifies Multiple KMS Relationship MPL Consumers,
+as they do not need to stand up LRU Cache of Hierarchy Keyrings.
```
texastony, line 166:
```suggestion
as they do not need to stand up a LRU Cache of Hierarchy Keyrings.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -144,7 +145,210 @@ is to make the Hierarchy Keyring
...
+
+Instead, they may maintain one CMC.
+They still create a Hierarchy Keyring instance per KMS Relationship,
+and they MUST use the correct Keyring to retrieve material
+from the Cache.
+
+But they only need not maintain the common cache.
```
texastony, line 173:
```suggestion
But they only need to maintain the common cache.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -144,7 +145,210 @@ is to make the Hierarchy Keyring
...
+
+Additionally, even though there is no CachingCMM in dafny right
+now, in addition to allowing a shared cache between multiple
+Hierarchy Keyrings, we want to create cache identifiers such
+that multiple CachingCMMs and multiple Hierarchy Keyrings can
+simultaneously use a single shared cache.
```
texastony, line 179: Issue: This paragraph is hard to read.

Can you re-write it? 
I think the summary is:
> In the future, the CachingCMM will be introduced to Crypto Tool's Dafny products.
The CachingCMM and the Hierarchy Keyring both consume CMCs.
Thus, it will be possible to provide a CMC to both a Hierarchy Keyring & a CachingCMM.
This design MUST consider how the cache entries of multiple Hierarchy Keyrings and CachingCMMs will be appropriately separated.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -144,7 +145,210 @@ is to make the Hierarchy Keyring
...
+  for multiple Hierarchy Keyrings and CachingCMMs.
+  For this, we need to strategically update the cache identifiers
+  for all materials ([background](#background)) in the CMC.
+  The [Issues and Alternatives section](#issues-and-alternatives)
+  discusses this.
+
+## Out of Scope
+
+- Supporting a shared CMC for multiple KMS relationships
+  in the **DB-ESDK** to fetch BeaconKeyMaterials.
```
texastony, line 193:
```suggestion
- Cache Across multiple Searchable Encryption Configurations, 
  or Searchable Encryption's caching allowing for  multiple KMS relationships.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -144,7 +145,210 @@ is to make the Hierarchy Keyring
...
+
+## Issues and Alternatives:
+
+Before looking at the issues, please take a quick look at
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula),
+[CachingCMM Cache Identifier formula](#cachingcmm-cache-identifier-formula), and
+[DB-ESDK Cache Identifier formula](#db-esdk-cache-identifier-formula) for more context.
```
texastony, line 200: Copy these over.
Once this work lands, the content will be different.
If you need the current formula's for this change doc to make sense, 
you will need to statically define them.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -210,29 +414,25 @@ it can be served without any KMS or DDB requests.
...
+- Minor update to the spec: The branch-key-version (which is the UUID) in the [Hierarchy Keyring Branch Key Materials (Decryption) cache identifier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/aws-kms/aws-kms-hierarchical-keyring.md#decryption-materials) is currently mentioned to be of size 36 bytes. However, the UUID is actually only 16 bytes of randomness. We get 36 bytes when we convert the 16 bytes of randomness in UUID to UTF8 encoding. Therefore, the spec should mention that the branch-key-version has size 16 bytes [[ref: Hierarchy Keyring Cache Identifier formula]](#hierarchy-keyring-cache-identifier-formula).
```
texastony, line 420: Suggestion: I know I used the term "entropy" and  "randomness", but technically, there is UUIDs are not a secure source of randomness.
 
```suggestion
- Minor update to the spec: The branch-key-version (which is the UUID) in the [Hierarchy Keyring Branch Key Materials (Decryption) cache identifier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/aws-kms/aws-kms-hierarchical-keyring.md#decryption-materials) is currently mentioned to be of size 36 bytes. However, the UUID is actually only 16 bytes. We get 36 bytes when we convert the 16 bytes in UUID to a UTF8 encoding. Therefore, the spec should mention that the branch-key-version has size 16 bytes [[ref: Hierarchy Keyring Cache Identifier formula]](#hierarchy-keyring-cache-identifier-formula).
```

texastony (changes requested):
This is a good start.
But you need to write more about how the inputs to the SHA are ensured to not collide.
Add content focusing on that.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: No, fixed it. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: I did not know that. Thanks for pointing this out.

RitvikKapila: Updated this in the doc wherever I mentioned DB-ESDK

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Okay, thanks for pointing this out!

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: I copied the CachingCMM cache identifier formula. I've tried to reference the current commit wherever possible. Thanks for pointing this out.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,454 @@
...
+The CachingCMM and the Hierarchy Keyring both consume CMCs.
+Thus, it will be possible to provide a CMC to both a
+Hierarchy Keyring & a CachingCMM.
+This cache identifier design MUST consider
+how the cache entries of multiple Hierarchy Keyrings
+and CachingCMMs will be appropriately separated.
+
+## Requirements
+
+- We must allow caching across KeyStores/KMS Clients/KMS Keys
```
texastony, line 186: nit: We have switched from KeyStore/keystore to Key Store. 
Two words.

RitvikKapila: Okay, thanks!

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+
+Note that there are two hashings taking place.
+For example in the Hierarchy Keyring Cache Identifier formula,
+one hash is for the branch-key-id and the other is hashing
+the entire appended string
+(a part of which is the hash of the branch-key-id)
+to get the final cache identifier.
+We'll call these internal and external hashes.
+
+#### Issue 1: How to update the cache identifier for the Hierarchy Keyring to allow a shared cache?
```
texastony, line 214: nit/formatting: you went from two `#` to four `#`, instead of 3 `#`.
I like neat hierarchies, 
but this 100% an author's choice.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+In the
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula)
+Encrypt, as discussed before, we will remove the len_branch_key.
+For Decrypt, we will remove the len_branch_key
+and use the branch-key-digest in place of the branch-key-id.
+The size of the inputs to the external hash for both these
+cases then becomes:
+
+- Encrypt: 17 (prefix) + 64 + 1 + 6 = 88 bytes
+- Decrypt: 17 + 64 + 1 + 36 = 118 bytes
```
texastony, line 383: Issue: You need to specify which of these bytes are constant and which are variable.
For example, the prefixes are constant.
But the branch-key-digest is variable.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+## Appendix
+
+## A1: Relating Cache Entries (Branch Keys) to KMS Relationships
+
+For some MPL Consumers,
+Cache Entries SHOULD only be served IF
+the request is being served under the same KMS Relationship,
+which is the most secure condition,
+as it prevents a Confused Deputy scenario
+(detailed in **Confused Deputy 1**).
```
texastony, line 472: Issue/Suggestion: Remove.
I never wrote this; it was on my TODO list.
Alternatively, if you want to write, you can. 
But I do not think it is needed today.
```suggestion
(detailed in **Confused Deputy 1**).
```

texastony (changes requested):
Progress!
Great Progress!
More than enough to move on to the implementation.
We may still need to refine the sections on the Cache Formulas. 

I need to look at the rendered view...

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+"Hard to Use".
+
+The objective, with these changes,
+is to make the Hierarchy Keyring
+"Easy to Use" in a multiple KMS Relationship
+environment.
+
+## Summary
+
+Relevant [sim](https://sim.amazon.com/issues/CrypTool-5311).
```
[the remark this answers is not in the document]
RitvikKapila: Great point. Thanks!

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: I had another heading there that I removed in the subsequent draft. Thanks for pointing this out.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+Note that according to the current
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula),
+if two KMS Relationships Hierarchy Keyrings are
+called with two different Key Stores but the same branch-key-id,
+there will be a collision (the branch-key-version will also
+have to co-incide in case of Decryption Branch Key Materials).
+
+##### Option 1 (Recommended): Append the keystore-id to the [Hierarchy Keyring cache identifier formula](#hierarchy-keyring-cache-identifier-formula).
+
+By appending the keystore-id to the
```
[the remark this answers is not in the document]
RitvikKapila: Great point. Thanks

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+use SHA512, but the Hierarchy Keyring cache identifier is
+truncated to 32 bytes.
+
+##### Option 1 (Recommended): Using SHA512 for both
+
+SHA512 has more number of bits to represent the output
+of the cache identifier strings. More information about
+why this is true based on the birthday bound can be found
+[here](#birthday-bound-for-sha512-and-sha384)
+This is pending [CryptoBR guidance](https://t.corp.amazon.com/P148158466/communication).
```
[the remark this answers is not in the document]
RitvikKapila: Yes, I'll do that. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: I'll remove this as there are more pertinent things to focus on.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+because they have different lengths. We just need
+to choose a good hashing algorithm which gives us a
+very large output space to project these strings to
+minimize collisions, and SHA512 will do the job.
+
+## Cache Identifier Formulae
+
+### Hierarchy Keyring Cache Identifier formula
+
+- [Encrypt](https://github.com/aws/aws-cryptographic-material-providers-library/blob/c07a51fc29ff70411f7573bca96d2a091db8c1ed/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsHierarchicalKeyring.dfy#L384) - SHA512(len_branch_key + SHA512(branch_key_id_utf8) + [0x00] + activeUtf8)[0..32] [[spec]](https://github.com/awslabs/aws-encryption-sdk-specification/tree/ffb2b0cc6a956b2cec3a33be3c3672605b6907fb/framework/aws-kms/aws-kms-hierarchical-keyring.md#encryption-materials)
```
nvobilis, line 405: len_branch_key is misleading. I believe it should refer to the length of the string branch-key-id, which is Uint32, as bytes

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+  - The SHA512(branch_key_id_utf8) is called the branch-key-digest
+- [Decrypt](https://github.com/aws/aws-cryptographic-material-providers-library/blob/c07a51fc29ff70411f7573bca96d2a091db8c1ed/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsHierarchicalKeyring.dfy#L725) - SHA512(len_branch_key + branch_key_id_utf8 + [0x00 as uint8] + UTF8.EncodeAscii(branchKeyVersion))[0..32] [[spec]](https://github.com/awslabs/aws-encryption-sdk-specification/tree/ffb2b0cc6a956b2cec3a33be3c3672605b6907fb/framework/aws-kms/aws-kms-hierarchical-keyring.md#decryption-materials)
```
nvobilis, line 407: [same](https://github.com/awslabs/aws-encryption-sdk-specification/pull/279/files#r1720116498)

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+Hierarchy Keyring cache identifier,
+we can ensure that there are no collisions between two identifiers
+from different Key Stores.
+More context on this can be found in the
+[Appendix section A1](#a1-relating-cache-entries-branch-keys-to-kms-relationships).
+
+##### Option 2: Append "unique numbers" to the cache identifiers of different keyrings
+
+This is not practical for a multi-threaded system.
+Also, we cannot keep track of these "unique numbers".
```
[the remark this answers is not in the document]
nvobilis: doesn't this option differ from option 1 because the keystore-id is "unique numbers" that we CAN track?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+[here](#birthday-bound-for-sha512-and-sha384)
+This is pending [CryptoBR guidance](https://t.corp.amazon.com/P148158466/communication).
+
+##### Option 2: Change the hashing algorithm to SHA384 in both
+
+SHA512 is the better choice as mentioned above.
+More information about this can be found
+[here in the appendix](#a3-information-about-sha384-and-length-extension-attacks),
+where I discuss length extension attacks and why
+SHA512 is still the better choice.
```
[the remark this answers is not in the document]
RitvikKapila: Yes, that's a good point. I've updated this now. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+customer provides.
+Additionally, if we're updating branch-key-id to
+branch-key-digest, we can remove the branch-key-id-length
+parameter in the formula and make it cleaner.
+This has CryptoBR's approval
+[[ticket]](https://issues.amazon.com/issues/P148158466).
+
+##### Option 2: Update the Encryption cache identifier to include branch-key-id instead of the branch-key-digest=SHA512(branch-key-id)
+
+Not hashing the branch keys will mean that the input
```
nvobilis, line 268: not hashing the branch key **ids**

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+#### Issue 1: How to update the cache identifier for the Hierarchy Keyring to allow a shared cache?
+
+Note that according to the current
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula),
+if two KMS Relationships Hierarchy Keyrings are
+called with two different Key Stores but the same branch-key-id,
+there will be a collision (the branch-key-version will also
+have to co-incide in case of Decryption Branch Key Materials).
+
+##### Option 1 (Recommended): Append the keystore-id to the [Hierarchy Keyring cache identifier formula](#hierarchy-keyring-cache-identifier-formula).
```
nvobilis, line 223: what if keystores are shared between keyrings?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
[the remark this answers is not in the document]
nvobilis: doesn't this option differ from option 1 because the keystore-id is "unique numbers" that we CAN track?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,438 @@
...
+The Hierarchy Keyring,
+and it's component the (Branch) Keystore,
+allow MPL Consumers to reduce their KMS Call volume
+by persisting KMS protected cryptographic materials into
+an available medium
+(currently, only a DynamoDB table is available as persistence medium).
+
+We call these cryptographic materials Branch Keys.
+
+However, an instance of the Hierarchy Keyring
```
nvobilis, line 92: does this mean, multiple keyrings may have the same KMS relationship?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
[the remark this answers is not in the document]
texastony: As I have read this:
- Option 1 is use the Key Store ID appended to the rest of the cache identifier.
- Option 2 is use a unique number appended to the rest of the cache identifier.

But, if the default value for Key Store ID is a UUID generated at local Key Store construction time,
how is this UUID practically different than a unique number?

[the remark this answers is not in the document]
RitvikKapila: Good point. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+called with two different Key Stores but the same branch-key-id,
+there will be a collision (the branch-key-version will also
+have to co-incide in case of Decryption Branch Key Materials).
+
+##### Option 1 (Recommended): Append the keystore-id to the [Hierarchy Keyring cache identifier formula](#hierarchy-keyring-cache-identifier-formula).
+
+By appending the keystore-id to the
+Hierarchy Keyring cache identifier,
+we can ensure that there are no collisions between two identifiers
+from different Key Stores.
```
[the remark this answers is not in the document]
RitvikKapila: Okay, sounds good. Thanks.

Adding a slightly modified version of this to the doc based on our slack conversation
'By adding Key Store ID, we let the customer control collisions.
I.e: they choose if different KMS Relationships CAN hit the same identifier or not.
They can also choose if different logical key stores or physical key stores can collide or not.
But by default, because each Key Store object will (likely) have a unique UUID,
there will be no collision.'

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Great idea. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,552 @@
...
+
+## A3: Information about SHA384 and Length Extension Attacks
+
+SHA384 computation speed is the same as that of SHA512 [[ref](https://news.ycombinator.com/item?id=27960348#:~:text=For%2064%2Dbit%20CPUs%20without,the%20algorithm%20is%20the%20same)].
+
+However, SHA384 is not prone to
+[Length Extension attacks](https://en.wikipedia.org/wiki/Length_extension_attack),
+unlike SHA512, which is a plus for SHA384.
+But, cache identifiers don't fall into the category
+of length extension attacks, which come into the picture
```
seebees, line 605: The reason cache identifiers may have issues is that if you can create a collision, then you would have 2 different branch key versions that are cached in the "same place".

Consider, you can select any branch key name you want.
So you select one such that the digest of your branch key will collide with mine.
You then make sure that you make a write request before me.
Now all my data will be encrypted with your branch key.

This is because the cache will be hydrated with you branch key,
and when the system looks to see if my branch key is cached
it will find an entry.

This can be mitigated in several ways.
But it is something to think about.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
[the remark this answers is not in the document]
RitvikKapila: Both of you are correct. Key Store IDs default to unique UUIDs that we CAN track. So we're gonna use Key Store ID (which default to a randomly generated UUID). In addition to that, a customer can set the Key Store ID to allow cache sharing between different KMS configurations. But by default, the Key Store IDs are unique, which is perfect for our use case.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: That's a good point. However, we're going to remove the length of the branch key ID from the cache identifier because it doesn't add anything.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Good catch. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Yes, you can create multiple keyrings with the same KMS relationships.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+| [AWS KMS Hierarchical Keyring](../../framework/aws-kms/aws-kms-hierarchical-keyring.md) |
+
+## Affected Implementations
+
+| Language | Version Introduced | Version Removed | Implementation                                                                                                                                                                                                                                                                                      |
+| -------- | ------------------ | --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Dafny    | 1.4.0              | n/a             | [AwsKmsHierarchicalKeyring.dfy](https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/src/Keyrings/AwsKms/AwsKmsHierarchicalKeyring.dfy)                                                          |
```
ShubhamChaturvedi7, line 28: Dafny 1.4.0 ?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+we need to allow caching across Key Stores/KMS Clients/KMS Keys.
+
+To facilitate Caching across Key Stores/KMS Clients/KMS Keys,
+we MUST break the Cryptographic Materials Cache (CMC)
+out of the Hierarchy Keyring.
+
+By allowing MPL Consumers to provide an already initialized CMC
+to the Hierarchy Keyring at construction,
+the CMC MAY cache Branch Keys protected by different
+KMS Relationships.
```
josecorella, line 166: Adding this cache MUST be an optional parameter, right? Otherwise current customers of the keyring will get compile errors saying, "must provide cache in initialization" or something like that. 
```suggestion
By allowing MPL Consumers to optionally provide an initialized CMC
to the Hierarchy Keyring at construction,
the CMC MAY cache Branch Keys protected by different
KMS Relationships.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+have to co-incide in case of Decryption Branch Key Materials).
+
+#### Option 1 (Recommended): Append the keystore-id to the [Hierarchy Keyring cache identifier formula](#hierarchy-keyring-cache-identifier-formula).
+
+By appending the [`keystore-id`](../../framework/branch-key-store.md#keystore-id) to the
+Hierarchy Keyring cache identifier,
+we can ensure that there are no collisions between two identifiers
+from different local Key Stores.
+
+Key Store IDs are UUIDs that are unlikely to collide by default.
```
texastony, line 233: nit: 
```suggestion
By default, Key Store IDs are UUIDs generated at object initialization; 
they are unlikely to collide unless they are manually set.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+But they only need to maintain the common cache.
+
+In the future, the CachingCMM will be introduced to
+Crypto Tool's Dafny products.
+The CachingCMM and the Hierarchy Keyring both consume CMCs.
+Thus, it will be possible to provide a CMC to both a
+Hierarchy Keyring & a CachingCMM.
+This cache identifier design MUST consider
+how the cache entries of multiple Hierarchy Keyrings
+and CachingCMMs will be appropriately separated.
```
josecorella, line 185: If this spec change is going to propose a change to caching cmm cache id formulas we should say that this cahnge also impacts the caching cmm up above or say that it is out of scope in the below section.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: There are two kinds of threats I can see here.

First, if both users use different key stores. If that is the case, the key store IDs will be different (unless the customer explicitly wants / sets them to be the same. 

Second, if they use the same key store, then, if I remember correctly, there can't be two branch key ids with the same name. The digest is a SHA512, so it is safe to assume that if the branch key ids are different, the digests will be different, given the large output space (2^512).

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Yes, you're right. It is an optional parameter that the customer can use.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
rishav-karanjit, line 233: User can also manually send the Key Store ID according to current spec. Are we defaulting it to UUID?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+Hash Functions is SHA-384.
+We believe they make this recommendation based on criteria
+other than collision resistance.
+
+<!-- TODO: Once CryptoBR has confirmed SHA-512 >> SHA-384 for
+this purpose, revise this section to state "We know" instead of "We believe" -->
+
+More information on length extension attacks
+in SHA384 and SHA512 can be found
+[here in the appendix](#a3-information-about-sha384-and-length-extension-attacks).
```
texastony, line 328: Nit: This "comes out of the blue".
What is missing, that you and I know, is that the SHA-512 is truncated to mitigate length extension attacks.

Maybe remove this sentence, 
but leave the appendix,
and add to the appendix that the original H-Keyring Formula 
mitigated length extension attacks,
but that mitigation is not needed.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+(a part of which is the hash of the branch-key-id)
+to get the final cache identifier.
+We'll call these internal and external hashes.
+
+### Issue 1: How to update the cache identifier for the Hierarchy Keyring to allow a shared cache?
+
+Note that according to the current
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula),
+if two KMS Relationships Hierarchy Keyrings are
+called with two different Key Stores but the same branch-key-id,
```
josecorella, line 222: we may want an example here. As I was reading; I said to myself, "ok, so? but then I was like right the KMS relationships above talk about clients with different regions and creds and so on"

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
rishav-karanjit: Tony's suggestion answered my question.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+[cachingCMM encrypt cache identifier](https://github.com/awslabs/aws-encryption-sdk-specification/tree/ffb2b0cc6a956b2cec3a33be3c3672605b6907fb/framework/caching-cmm.md#appendix-a-cache-entry-identifier-formulas)
+has two different formulae based on if we
+specify the algorithm suite or not).
+
+In the
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula)
+Encrypt, as discussed before, we will remove the len_branch_key.
+For Decrypt, we will remove the len_branch_key
+and use the branch-key-digest in place of the branch-key-id.
+We also add the keystore-id, which is a UUID,
```
texastony, line 404: issue: keystore-id defaults to a UUID, but it can be user input... so it can be `n` length.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+the CMC MAY cache Branch Keys protected by different
+KMS Relationships.
+
+This simplifies Multiple KMS Relationship MPL Consumers,
+as they do not need to stand up a LRU Cache of Hierarchy Keyrings.
+
+Instead, they may maintain one CMC.
+They still create a Hierarchy Keyring instance per KMS Relationship,
+and they MUST use the correct Keyring to retrieve material
+from the Cache.
```
ShubhamChaturvedi7, line 174: Does this mean we are exposing an interface to accept a Cache while the internal working remains (mostly) the same? Or are we changing how materials are cached as well (excluding the identifier change).

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
ShubhamChaturvedi7, line 233: I thought they were the name of the DDB Table. How this UUID is generated?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+For instance, they can choose if different
+KMS Relationships CAN hit the same identifier or not.
+They can also choose if different logical key stores
+or physical key stores can collide or not.
+But by default, because each Key Store object will have a unique UUID,
+there will be no collision.
+More context on this can be found in the
+[Appendix section A1](#a1-relating-cache-entries-branch-keys-to-kms-relationships).
+
+#### Option 3: Don't change anything.
```
ShubhamChaturvedi7, line 246: Where's Option 2?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+Note that according to the current
+[Hierarchy Keyring Cache Identifier formula](#hierarchy-keyring-cache-identifier-formula),
+if two KMS Relationships Hierarchy Keyrings are
+called with two different Key Stores but the same branch-key-id,
+there will be a collision (the branch-key-version will also
+have to co-incide in case of Decryption Branch Key Materials).
+
+#### Option 1 (Recommended): Append the keystore-id to the [Hierarchy Keyring cache identifier formula](#hierarchy-keyring-cache-identifier-formula).
+
+By appending the [`keystore-id`](../../framework/branch-key-store.md#keystore-id) to the
```
josecorella, line 228: hmm what if I create my keystore with a custom keystore-id?
Since it could be supplied this would or wouldn't work?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
josecorella, line 246:
```suggestion
#### Option 2: Don't change anything.
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Ohh yes you're right.

Because we want the cache identifier to be of uniform length, the best way to move forward would be to hash the Key Store ID.

I'll update this in the doc after the meeting.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+the [Hierarchy Keyring Branch Key Materials (Decryption) cache identifier](https://github.com/awslabs/aws-encryption-sdk-specification/tree/ffb2b0cc6a956b2cec3a33be3c3672605b6907fb/framework/aws-kms/aws-kms-hierarchical-keyring.md#decryption-materials)
+has branch-key-id appended in the cache identifier formula and the
+[Branch Key Materials (Encryption) cache identifier](https://github.com/awslabs/aws-encryption-sdk-specification/tree/ffb2b0cc6a956b2cec3a33be3c3672605b6907fb/framework/aws-kms/aws-kms-hierarchical-keyring.md#encryption-materials)
+has branch-key-digest=SHA512(branch-key-id). Here we're talking
+about the internal hash.
+
+#### Option 1 (Recommended): Update the Decryption cache identifier to include branch-key-digest=SHA512(branch-key-id) instead of the branch-key-id
+
+This will make both the Hierarchy Keyring Encryption and
+Decryption cache identifiers more coherent.
```
lucasmcdonald3, line 265: nit: Do you mean consistent? I think the formulas are coherent but not consistent.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+
+### Issue 4: Should we add a prefix to all cache identifiers in Hierarchy Keyrings and CachingCMMs?
+
+Here, by prefix we mean, adding the UTF8 encoding
+of the string 'hierarchy_keyring' as a
+prefix to a hierarchy keyring cache identifier,
+and adding encoding of 'caching_cmm' as a prefix to a
+CachingCMM cache identifier.
+
+#### Option 1 (Recommended): Yes
```
ShubhamChaturvedi7, line 344: I feel this is trivial. If you are not trusting your hash function then there are bigger issues, which is beyond adding a prefix.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
nvobilis: so then wouldn't multiple keyrings (sharing the same keystore) create cache IDs with collision?

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+Hierarchy Keyring and a CachingCMM cache identifier.
+Adding prefixes adds another layer of distinction between them.
+
+## How can we ensure cache identifiers are collision resistent?
+
+To answer this question, I will use two approaches.
+
+### Derivation from assumptions
+
+For the HKeyring and CachingCMM, ensuring that all cache
```
josecorella, line 367: nit: I think this is the first time you directly reference HKeyring, which is fine, but when you first mention the Hierarchy Keyring add (HKeyring) afterwards to make it clear we are shortening it.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
texastony: Nit: One threat, two exploits/mitigations, depending on how you look at it.

But I think Ryan is on to something;
we need to describe the threat in this doc.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Good point. I'll do that. Thanks.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: We are updating the spec to update the cachingCMM cache identifiers.

Currently, we're only focussing on making the changes to the MPL. Eventually, everything is going to be moved to Dafny anyway.

If a case can be made for other native libraries, we can incorporate that.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
```diff
@@ -0,0 +1,613 @@
...
+the MPL Consumer SHOULD recreate
+
+> a KMS Client is created that respects relationship `X`,
+> which is then used to create Key Store Identified as `X`,
+> and the Key Store is used to create a Hierarchical Keyring
+> that uses the common Cache.
+
+If the Cache Entry is still valid
+(the TTL has not expired),
+it can be served without any KMS or DDB requests.
```
ShubhamChaturvedi7, line 587: Do we have any concrete example? Anything that has to do with GC gets tricky and can be source of hard to debug bugs.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
texastony: Maybe? 
Alternatively, we could "invent" a new field for the Key Store object;
key store digest.

The Key Store digest is either:
- UUID, if no Key Store ID is given
- SHA(Key Store ID) if Key Store ID is given

If we can find a SHA that gives 16 bytes, you end up with 16 bytes either way.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Update: Clarify between accidental collisions and explicit cache hits.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Moving to using partitionId instead of keyStoreId

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Discussed - This is the MPL version

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Updated the cache identifiers. This shouldn't be an issue now.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Fixed

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: We moved to using partition Id instead of Key Store Id. But both these are modeled in the same way. If the partition Id is provided by the customer, we set the partition Id to utf8 encoding of the input. Otherwise, we set it to a randomly generated 16 byte UUID.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: First. The cache interface does NOT change. We are only changing the way we take a shared cache as input.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Update: Native library cache identifiers can NEVER collide with the new proposed cache identifiers.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
texastony, line 1: I am not reviewing the change doc today.

framework/aws-kms/aws-kms-hierarchical-keyring.md
texastony, line 1: Blocking:
- Bump the Version on line 8 to 0.2.0.
- Add a Changelog entry

texastony (changes requested):
Blocking on changes need in framework.
Will review `change.md` another time.

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: This should be fixed with the updated with the most recent cache identifiers: `partitionId + logicalKeyStoreName + branchKeyId`

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Thanks. Updated.

changes/2024-09-13_cache-across-hierarchical-keyrings/background.md
```diff
@@ -0,0 +1,415 @@
...
+### Issue 1: How to update the cache identifier for the Hierarchical Keyring to allow a shared cache?
+
+Note that according to the current
+[Hierarchical Keyring Cache Identifier formula](#hierarchical-keyring-cache-identifier-formula),
+if two KMS Relationships Hierarchical Keyrings are
+called with two different Key Stores but the same branch-key-id,
+there will be a collision (the branch-key-version will also
+have to co-incide in case of Decryption Branch Key Materials).
+
+#### _Option 1 (Recommended): Adding a Resource ID, Scope ID, Partition ID, and Resource Suffix (recommended)_
```
josecorella, line 71: I think this title is out of date

changes/2024-09-13_cache-across-hierarchical-keyrings/background.md
```diff
@@ -0,0 +1,415 @@
...
+  A Hex value that indicates
+  if an element is used for Encryption, Decryption,
+  Searchable Encryption, or some other future
+  purpose.
+
+  - Encrypt : `0x01` (0001)
+  - Decrypt : `0x02` (0010)
+  - Searchable Encryption : `0x03` (0011)
+
+- **Partition ID:**
```
josecorella, line 94: let's flip this around and talk about the default behavior first

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,188 @@
...
+we need to allow caching across Key Stores/KMS Clients/KMS Keys.
+
+To facilitate caching across Key Stores/KMS Clients/KMS Keys,
+we MUST break the Cryptographic Materials Cache (CMC)
+out of the Hierarchical Keyring.
+
+By allowing MPL Consumers to optionally provide an initialized shared CMC
+to the Hierarchical Keyring at construction,
+the CMC MAY cache Branch Keys protected by different
+KMS Relationships.
```
josecorella, line 67: let's add a note here saying that doing this allows any keyring that has access to the cache may be able to use materials that they may or may not have direct access to.

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,188 @@
...
+how the cache entries of multiple Hierarchical Keyrings
+and CachingCMMs will be appropriately separated.
+
+## Out of Scope
+
+- Supporting a shared Cache in the DB-ESDK to fetch
+  BeaconKeyMaterials across multiple Searchable
+  Encryption Configurations, or across multiple KMS Relationships.
+
+## Motivation
```
josecorella, line 94: feels like motivation should be before the summary

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,188 @@
...
+and clearly makes the Hierarchical Keyring,
+in these conditions,
+"Hard to Use".
+
+The objective, with these changes,
+is to make the Hierarchical Keyring
+"Easy to Use" in a multiple KMS Relationship
+environment.
+
+## Security Implications
```
josecorella, line 161: let's expand on the implications of the same partitionID across keyrings, bypassing access control and all that jazz

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,188 @@
...
+to be the same for two Hierarchical Keyrings if and only if they want them to share cache entries.
+
+## Operational Implications
+
+This change will allow customers to share an already initialized cache across multiple
+Hierarchical Keyrings to facilitate caching across Key Stores/KMS Clients/KMS Keys.
+
+## Examples
+
+As part of this change, we will add two examples, one each in ESDK .NET and ESDK JAVA
```
josecorella, line 187: nit:
```suggestion
As part of this change, we will add two examples, one each in ESDK .NET and ESDK Java
```

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -47,18 +53,40 @@ On initialization, the caller:
...
 
 ### Cache Limit TTL
 
 The maximum amount of time in seconds that an entry within the cache may be used before it MUST be evicted.
 The client MUST set a time-to-live (TTL) for [branch key materials](../structures.md#branch-key-materials) in the underlying cache.
 This value MUST be greater than zero.
 
+### Cache Type
+
+Sets the cache bound OR the type of cache scoped (created & bound) to the Hierarchical Keyring.
```
josecorella, line 72: this sentence is confusing

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -47,18 +53,40 @@ On initialization, the caller:
...
+That is, when it is garbage collected or if the cache is bound to other Cryptographic Material Providers.
+If any other type in `CacheType` is provided, the Hierarchical Keyring
+will initialize a cache of that type, to be used with only this Hierarchical Keyring.
+If not set, a `DefaultCache` is initialized to be used with only this Hierarchical Keyring with entryCapacity = 1000.
+
+### Partition ID
+
+A string that is used to avoid collisions with other Hierarchical Keyrings.
+If this parameter is not set, the Hierarchical Keyring MUST set a partition ID
+that uniquely identifies the respective Hierarchical Keyring.
```
josecorella, line 83: MUST be set to what?

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -294,66 +334,204 @@ in order to compute the [cache entry identifier](../cryptographic-materials-cach
...
-- MUST be the constant UTF8 encoded string "ACTIVE"
+#### Resource Suffix
 
-| Field                    | Length (bytes) | Interpreted as |
-| ------------------------ | -------------- | -------------- |
-| Length of branch-key-id  | 3              | UInt8          |
-| SHA-512(branch-key-id)   | 64             | bytes          |
-| Null Byte                | 1              | `0x00`         |
-| Constant string "ACTIVE" | 6              | UTF-8 Encoded  |
+There are, at this time, 2 resource suffixes for the Hierarchical Keyring:
```
josecorella, line 373:
```suggestion
There are two resource suffixes for the Hierarchical Keyring:
```

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -294,66 +334,204 @@ in order to compute the [cache entry identifier](../cryptographic-materials-cach
...
+    + NULL_BYTE
+    + logical-key-store-name
+    + NULL_BYTE
+    + branch-key-id
+    + NULL_BYTE
+    + branch-key-version
+)
 ```
 
+## Shared Cache Considerations
```
josecorella, line 492: Let's also add here the detail about the shared access to materials

josecorella (changes requested):

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Should be clarified now with the latest updates.

framework/caching-cmm.md
```diff
@@ -197,73 +199,215 @@ to the denote the function that,
...
+Searchable Encryption : 0x03 (0011)
+```
+
+#### Partition ID
+
+Partition ID of the Caching CMM.
+
+#### Resource Suffix
+
+There are, at this time, 3 resource suffixes for the Caching CMM:
```
RitvikKapila, line 229:
```suggestion
There are 3 resource suffixes for the Caching CMM:
```

changes/2024-TBD_cache-across-hierarchy-keyrings/change.md
RitvikKapila: Resolving this. Removed the design doc and created a background / change doc. I don't think we need this examples there.

changes/2024-09-13_cache-across-hierarchical-keyrings/background.md
RitvikKapila: Sounds good. Thanks for the suggestion.

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
RitvikKapila: If you read the whole doc then yes, you're right. But looking at the previous change docs, Summary is at the top because that is the most important thing anyone is going to read.

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
RitvikKapila: Great idea. Updated the spec. Thanks!

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
RitvikKapila: I added a lot of stuff here. Let me know if you think we should cut it down.

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Good point. It sounded different when this was first written. Thanks for bringing this up. This same documentation is in the smithy model for keyrings where the cache input is defined for the Hierarchical Keyring. I updated that and the spec here. Let me know if this updated explanation makes more sense.

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Updated. Thanks!

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: I added a brief note mentioning this here. Thanks!

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+- Same Partition ID
+- Same Logical Key Store Name of the Key Store for the Hierarchical Keyring
+- Same Branch Key ID
+
+then the Hierarchical Keyrings WILL share cache entries in the `Shared` Cache.
+
+This means that the branch-key used and cached by one Hierarchical Keyring can be
+used by the other Hierarchical Keyring (within a TTL), effectively by-passing
+KMS Access Control. We recommend evaluating your threat model carefully, to
+understand and accept / mitigate this risk. By default, each Hierarchical Keyring's
```
josecorella, line 188:
```suggestion
understand and mitigate this risk yourself. By default, each Hierarchical Keyring's
```

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -294,66 +341,213 @@ in order to compute the [cache entry identifier](../cryptographic-materials-cach
...
+Therefore, there are two important parameters that users need to carefully set while providing the shared cache:
+
+### Partition ID
+
+Partition ID is an optional parameter provided to the Hierarchical Keyring input,
+which distinguishes Cryptographic Material Providers (i.e: Hierarchical Keyrings) writing to a cache.
+
+- (Default) If the Partition ID is not set by the user, it is initialized as a random 16-byte UUID, which makes
+  it unique for every Hierarchical Keyring. In this case, two Hierarchical Keyrings (or another Material Provider)
+  CANNOT share the same cache entries in the cache.
```
josecorella, line 526:
```suggestion
- (Default) A a random 16-byte UUID, which makes
  it unique for every Hierarchical Keyring. In this case, two Hierarchical Keyrings (or another Material Provider)
  CANNOT share the same cache entries in the cache.
```

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
```diff
@@ -0,0 +1,216 @@
...
+## Out of Scope
+
+- Supporting a shared Cache in the DB-ESDK to fetch
+  BeaconKeyMaterials across multiple Searchable
+  Encryption Configurations, or across multiple KMS Relationships.
+
+## Motivation
+
+The Hierarchical Keyring,
+and it's component the (Branch) Key Store,
```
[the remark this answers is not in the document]
RitvikKapila: Good point. Thanks!

changes/2024-09-13_cache-across-hierarchical-keyrings/change.md
RitvikKapila: Thanks!

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -47,18 +53,43 @@ On initialization, the caller:
...
+cache of that type, to be used with only this Hierarchical Keyring. If not set, a `DefaultCache` is initialized
+to be used with only this Hierarchical Keyring with `entryCapacity` = 1000.
+
+### Partition ID
+
+An optional string that uniquely identifies the respective Hierarchical Keyring
+and is used to avoid collisions with other Hierarchical Keyrings.
+
+PartitionId can be a String provided by the user, and in this case it MUST be interpreted as the bytes of
+UTF-8 Encoding of the String.
```
josecorella, line 85:
```suggestion
PartitionId can be a string provided by the user. If provided it MUST be interpreted as UTF8 bytes.
```

framework/aws-kms/aws-kms-hierarchical-keyring.md
```diff
@@ -294,66 +341,213 @@ in order to compute the [cache entry identifier](../cryptographic-materials-cach
...
+  they CAN share the same cache entries in the cache.
+- If the Partition ID is set by the user and is different for two Hierarchical Keyrings (or another Material Provider),
+  they CANNOT share the same cache entries in the cache.
+
+### Logical Key Store Name
+
+> Note: Users MUST NEVER have two different physical Key Stores with the same Logical Key Store Name.
+
+Logical Key Store Name is set by the user when configuring the Key Store for
+the Hierarchical Keyring. This is a logical name for the branch key store.
```
josecorella, line 537:
```suggestion
the Hierarchical Keyring. This is a logical name for the key store.
```

josecorella (approved):
looks good

framework/aws-kms/aws-kms-hierarchical-keyring.md
RitvikKapila: Thanks for pointing this out! Fixed.

PR #283, opened 2024-09-11
fix: Names and clarification

seebees:
The encyrption context on hierarchical materials
is all the encryption context,
not only any custom values set by the customer
on branch key creation.

# Check any applicable:


lucasmcdonald3 (approved):

texastony (commented):
I am not going to look right now, but I do not know if this was the right spec update.

Did anyone run a grep to see where `#encryption-context-3` is used?

PR #290, opened 2025-03-13
fix: Update symmetric signing keys

seebees:
0 and `null` are both OK.
This is how the Java type conversion
works for the MPL Dafny.

See: https://github.com/aws/aws-cryptographic-material-providers-library/pull/1328

# Check any applicable:


ajewellamz (approved):
LGTM

PR #267, opened 2024-04-19
chore: update for mrk support in keystore

ajewellamz:
_Issue #, if available:_

_Description of changes:_

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Check any applicable:

- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #271, opened 2024-05-19
feat(Keystore): Introduce additional KMSConfiguration options

texastony:
_Issue #, if available:_

_Description of changes:_
This PR is logically equivalent #270  .
However, @seebees and @texastony have agreed to merge this PR 
and leave #270 open so we can resolve the open issues in it
at a later date.

Thus, this PR will be merged to master when
https://github.com/aws/aws-cryptographic-material-providers-library/pull/316
is merged to mainline.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Check any applicable:

- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #291, opened 2025-05-07
fix: Branch key creation uses unchanged input

seebees:
Discovery and chainging the region of a KMS key
on decrypt do not mutate the customer input.
This aligns with customer intent.
Further the behavior of the code is to use the provided key directly

#267 underspecified how branch keys should be created so this PR rolls back that change.

# Check any applicable:


ajewellamz (approved):
LGTM

framework/branch-key-store.md
```diff
@@ -9,6 +9,8 @@
 
 ### Changelog
 
+- 0.7.1
+  - Branch key creation only uses customer input (pre 0.5 behavior)
```
rishav-karanjit, line 13: Question: The below changes looks like fixes that adds missing parameter thats send to KMS. 

I am trying to understand what does `Branch key creation only uses customer input` and what does `(pre 0.5 behavior)` mean?

rishav-karanjit: I think I got what `Branch key creation only uses customer input` means. It mean `Key ARN` is a customer input but not sure about `(pre 0.5 behavior)`

seebees: I'm looking more at the changes. I think I'll drop it
```suggestion
  - Branch key creation only uses customer input
```

The idea was that looking at the version changes, I thought that this would be in the 0.5 version change (#271). But when looking more at that PR, I think I was less right.

ajewellamz (approved):

rishav-karanjit (approved):
LGTM

PR #295, opened 2025-06-12
chore: clarify and update the storm tracking cache

ajewellamz:
_Issue #, if available:_

_Description of changes:_

# Check any applicable:

### DIFF
--- a/framework/storm-tracking-cryptographic-materials-cache.md
+++ b/framework/storm-tracking-cryptographic-materials-cache.md
@@ -7,6 +7,10 @@
 
 ### Changelog
 
+- 0.2.0
+  - Add Time Units
+  - Clarify Inflight TTL and Grace Interval
+
 - 0.1.0
   - Initial record
   - [Thread Safe Cache](../changes/2023-06-19_thread_safe_cache/change.md)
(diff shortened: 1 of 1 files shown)


framework/storm-tracking-cryptographic-materials-cache.md
```diff
@@ -53,24 +63,41 @@ attempts will be made to refresh the cache.
...
+
+The Grace Interval should be larger than the longest time expected for a request to the underlying cryptographic materials provider.
 
 ### FanOut
 
 A number (at least 1, default 20).
 
 The maximum number of individual keys for which lookups can be in flight.
 
+Other requests are blocked until the number of outstanding requests drops below this number,
```
seebees, line 87: I think you should clarify what we mean by requests here. Because if the request is for exactly the same ID, then it is not blocked. it would be returned. This has to do with requests to different ids yes?

framework/storm-tracking-cryptographic-materials-cache.md
```diff
@@ -53,24 +63,41 @@ attempts will be made to refresh the cache.
...
+
 ### Inflight TTL
 
 A number (at least 1, default 20).
 
-An entry that has been in flight for this long is no longer considered in flight.
+A request that has been in flight for this long returns an "InFlightTTL exceeded" error.
+
+If a thread has been sleeping because other threads are allegedly working on the same key,
+give up and return an error after this amount of time.
```
seebees, line 97:
```suggestion
it will give up and return an error after this amount of time.
```

seebees (commented):
small nit

PR #300, opened 2025-07-18
chore: encryption-context changes

rishav-karanjit:
_Issue #, if available:_

_Description of changes:_
This PR is getting merged to `rishav/hv-2-spec`. This PR addresses comments in [PR 300](https://github.com/awslabs/aws-encryption-sdk-specification/pull/300) that is related to encryption context only. 

Note: If `encryption-context` is mentioned in the spec then do look into the input of the operation as the operation input will be specifying `encryption-context` which will clear up the context of `encryption-context`.

# Check any applicable:

### DIFF
--- a/framework/branch-key-store.md
+++ b/framework/branch-key-store.md
@@ -745,13 +745,14 @@ The Branch Key Context:
 
 - MUST have a `branch-key-id` key who's value MUST not be an empty string
 - MUST have a `type` key who's value MUST not be an empty string
-- MUST have a `create-time` key in ISO 8601 format in UTC
+- MUST have a `create-time` key with a value in ISO 8601 format in UTC
 - MUST have a `tablename` key who's value is the logicalKeyStoreName
 - MUST have a `kms-arn` key who's value is valid KMS ARN
 - MUST have a `hierarchy-version` key who's value is either "1" or "2"
 - MUST NOT have a `enc` key
 - MAY have one or more keys prefixed with `aws-crypto-ec:` which is the encyption context send by the customer.
   The `aws-crypto-ec:` prefix is prepended by the library
+- MUST NOT have any other keys apart from the ones mentioned above if `hierarchy-version` is "2"
 
 ### ACTIVE Branch Key Context
 
--- a/framework/branch-key-store.md
+++ b/framework/branch-key-store.md
@@ -518,7 +518,7 @@ List of TransactWriteItem:
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
     - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
-    - Every key-value pair of the [encryption context](#encryption-context) that is associated with the branch key
+    - Every key-value pair of the [encryption context](./structures.md#encryption-context) that is associated with the branch key
       MUST be added with an Attribute Name of `aws-crypto-ec:` + the Key and Attribute Value (S) of the value.
   - ConditionExpression: `attribute_not_exists(branch-key-id)`
   - TableName: the configured Table Name
@@ -530,7 +530,7 @@ List of TransactWriteItem:
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
     - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
-    - Every key-value pair of the custom [encryption context](#encryption-context) that is associated with the branch key
+    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context) that is associated with the branch key
       MUST be added with an Attribute Name of `aws-crypto-ec:` + the Key and Attribute Value (S) of the value.
   - ConditionExpression: `attribute_not_exists(branch-key-id)`
   - TableName: the configured Table Name
@@ -542,7 +542,7 @@ List of TransactWriteItem:
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
     - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
-    - Every key-value pair of the custom [encryption context](#encryption-context) that is associated with the branch key
+    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context) that is associated with the branch key
       MUST be added with an Attribute Name of `aws-crypto-ec:` + the Key and Attribute Value (S) of the value.
   - ConditionExpression: `attribute_not_exists(branch-key-id)`
   - TableName is the configured Table Name
@@ -624,18 +624,18 @@ Otherwise, this operation MUST yield an error.
 The operation MUST use the configured `KMS SDK Client` to authenticate the value of the keystore item.
 
 This operation MUST call AWS DDB `GetItem`.
-Every attribute except `enc` on the AWS DDB response item MUST be converted to a set of key value pairs (a Map)
-which is the [AWS KMS Encryption Context](#aws-kms-encryption-context).
+Every attribute except `enc` on the AWS DDB response item MUST be converted to a set of key value pairs
+which is the [branch key context](#branch-key-context).
 
 The operation MUST call [AWS KMS API ReEncrypt](https://docs.aws.amazon.com/kms/latest/APIReference/API_ReEncrypt.html)
 with a request constructed as follows:
 
-- `SourceEncryptionContext` MUST be the [AWS KMS Encryption Context](#aws-kms-encryption-context) constructed above
+- `SourceEncryptionContext` MUST be the [branch key context](#branch-key-context) constructed above
 - `SourceKeyId` MUST be [compatible with](#aws-key-arn-compatibility) the configured KMS Key in the [AWS KMS Configuration](#aws-kms-configuration) for this keystore.
 - `CiphertextBlob` MUST be the `enc` attribute value on the AWS DDB response item
 - `GrantTokens` MUST be the configured [grant tokens](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#grant_token).
 - `DestinationKeyId` MUST be [compatible with](#aws-key-arn-compatibility) the configured KMS Key in the [AWS KMS Configuration](#aws-kms-configuration) for this keystore.
-- `DestinationEncryptionContext` MUST be the [AWS KMS Encryption Context](#aws-kms-encryption-context) constructed above
+- `DestinationEncryptionContext` MUST be the [branch key context](#branch-key-context) constructed above
 
 #### Authenticating a Branch Keystore item for item with `hierarchy-version` v2
 
@@ -672,7 +672,7 @@ The operation MUST decrypt the branch key according to the [AWS KMS Branch Key D
 If the branch key fails to decrypt, GetActiveBranchKey MUST fail.
 
 This GetActiveBranchKey MUST construct [branch key materials](./structures.md#branch-key-materials)
-according to [Branch Key Materials From Authenticated Encryption Context](#branch-key-materials-from-authenticated-encryption-context).
+according to [Branch Key Materials From Authenticated Branch Key Context](#branch-key-materials-from-authenticated-branch-key-context).
 
 This operation MUST return the constructed [branch key materials](./structures.md#branch-key-materials).
 
@@ -698,7 +698,7 @@ The operation MUST decrypt the branch key according to the [AWS KMS Branch Key D
 If the branch key fails to decrypt, this operation MUST fail.
 
 This GetBranchKeyVersion MUST construct [branch key materials](./structures.md#branch-key-materials)
-according to [Branch Key Materials From Authenticated Encryption Context](#branch-key-materials-from-authenticated-encryption-context).
+according to [Branch Key Materials From Authenticated Branch Key Context](#branch-key-materials-from-authenticated-branch-key-context).
 
 This operation MUST return the constructed [branch key materials](./structures.md#branch-key-materials).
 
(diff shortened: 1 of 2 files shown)


framework/structures.md
```diff
@@ -467,7 +468,7 @@ the beacon key can be discarded.
 
 ##### Encryption Context
 
-The [custom encryption context](#encryption-context) associated with this beacon key.
+The [encryption context](#encryption-context) associated with this beacon key.
```
seebees, line 471: Why is the encryption context of the branch key prepended but the beacon key is not?

framework/structures.md
```diff
@@ -414,7 +414,8 @@ This value MUST be a version 4 [UUID](https://www.ietf.org/rfc/rfc4122.txt).
 
 ##### Encryption Context
 
-The [custom encryption context](#encryption-context) associated with this branch key.
+The [encryption context](./structures.md#encryption-context) is a key-value pair provided by users and is associated with the branch key.
+The key of this encryption context is alway prepended with `aws-crypto-ec:`.
```
seebees, line 418: Are we ok with this? This is very technically a breaking change right?

framework/branch-key-store.md
```diff
@@ -870,36 +846,36 @@ A branch key record MUST include the following key-value pairs:
 1. `hierarchy-version`: Version of the hierarchical keyring;
    represented as [AWS DDB Number](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes)
 
-A branch key record MAY include [custom encryption context](#custom-encryption-context) key-value pairs.
-These attributes MUST always be prefixed with `aws-crypto-ec:` regardless of the item's `hierarchy-version`.
+A branch key record MAY include [encryption context](./structures.md#encryption-context) key-value pairs.
+These attributes MUST be always prefixed with `aws-crypto-ec:` regardless of the item's `hierarchy-version`.
```
texastony, line 850:
```suggestion
These attributes MUST be prefixed with `aws-crypto-ec:` regardless of the item's `hierarchy-version`.
```

framework/branch-key-store.md
```diff
@@ -534,7 +534,7 @@ List of TransactWriteItem:
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
     - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
-    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context-3) that is associated with the branch key
+    - Every key-value pair in the [encryption context](./structures.md#encryption-context) provided by users and is associated with the branch key
```
seebees, line 537: This is the creation.
So I think you want to reference "The original input `encryption-context`"

framework/branch-key-store.md
```diff
@@ -522,7 +522,7 @@ List of TransactWriteItem:
...
```
seebees, line 525: This is the creation.
So I think you want to reference "The original input `encryption-context`"

framework/branch-key-store.md
```diff
@@ -870,36 +846,36 @@ A branch key record MUST include the following key-value pairs:
...
+The [encryption context](./structures.md#encryption-context) is stored as map of UTF8 Encoded bytes.
 
-For every key in the [encryption context](./structures.md#encryption-context-3)
+For every key in the [encryption context](./structures.md#encryption-context)
 the string `aws-crypto-ec:` + the UTF8 decode of this key
 MUST exist as a key in the authenticated encryption context.
-Also, the value in the [encryption context](./structures.md#encryption-context-3) for this key
+Also, the value in the [encryption context](./structures.md#encryption-context) for this key
 MUST equal the value in the authenticated encryption context
 for the constructed key.
```
texastony, line 880: Suggestion: "constructed" is vague; prefixed is more explicit.
```suggestion
for the prefixed key.
```

framework/branch-key-store.md
```diff
@@ -601,8 +601,8 @@ List of TransactWriteItem:
     - “enc” (B): the wrapped DECRYPT_ONLY Branch Key `CiphertextBlob` from the KMS operation
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
-    - “hierarchy-version” (N): 1
-    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context-3) that is associated with the branch key
+    - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
+    - Every key-value pair in the [encryption context](./structures.md#encryption-context) provided by users and is associated with the branch key
```
seebees, line 605: This is version key, so it is slightly different.

What is the "encryption context provided by users" at this point? I think we want to be careful at this point. Can we just refer to the branch key context here? And say "Just write the whole context like this"? Because that will already have anything prepended correct?

framework/branch-key-store.md
```diff
@@ -613,8 +613,8 @@ List of TransactWriteItem:
     - “enc” (B): wrapped ACTIVE Branch Key `CiphertextBlob` from the KMS operation
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
-    - “hierarchy-version” (N): 1
-    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context-3) that is associated with the branch key
+    - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
+    - Every key-value pair in the [encryption context](./structures.md#encryption-context) provided by users and is associated with the branch key
```
seebees, line 617: same

framework/branch-key-store.md
```diff
@@ -652,7 +652,7 @@ This operation MUST call AWS DDB `GetItem`.
 Every attribute except `enc` on the AWS DDB response item MUST be converted to a set of key value pairs (a Map)
 which is the [branch key context](#branch-key-context).
 Every key in the constructed branch key context except tableName MUST exist as a string attribute in the AWS DDB response item.
-The [encryption context](#encryption-context) MUST be built from branch key context by extracting keys with prefix `aws-crypto-ec:` and then dropping the prefix.
+The [AWS KMS Encryption Context](#aws-kms-encryption-context) MUST be built from branch key context by extracting keys with prefix `aws-crypto-ec:` and then dropping the prefix.
```
seebees, line 655: I think you want a very similar statement that describes how to create the branch key context on Create.

framework/branch-key-store.md
```diff
@@ -779,34 +779,10 @@ In addition to the [branch key context](#branch-key-context):
...
-If the `hierarchy-version` is v1, Beacon Encryption Context is same as [Beacon Key Branch Key Context](#beacon-key-branch-key-context)
-
-### Custom Encryption Context
-
-If custom [encryption context](./structures.md#encryption-context)
-is associated with the branch key these values MUST be added to the AWS KMS encryption context.
-To avoid name collisions each added attribute from the custom [encryption context](./structures.md#encryption-context-3)
-MUST be prefixed with `aws-crypto-ec:`.
-Across all versions of a Branch Key, the custom encryption context MUST be equal.
+If the `hierarchy-version` is v2, AWS KMS encryption context MUST be the [encryption context](./structures.md#encryption-context) sent by users without any transformation. This requirement applies regardless of the Branch Key Item's `type`.
```
seebees, line 785: You have this nice section here: #encryption-context-from-authenticated-branch-key-context. It think that this better describes the relationship than "send by the users".

framework/branch-key-store.md
```diff
@@ -844,7 +820,7 @@ the keystore operation MUST call with a request constructed as follows:
   If the KMS Configuration is MRDiscovery, `KeyId` MUST be the `kms-arn` attribute value of the AWS DDB response item, with the region replaced by the configured region.
   Otherwise, it MUST BE the Keystore's configured KMS Key.
 - `CiphertextBlob` MUST be the `enc` attribute value on the AWS DDB response item
-- `EncryptionContext` MUST be the [encryption context](#encryption-context) constructed above
+- `EncryptionContext` MUST be the [AWS KMS Encryption Context](#aws-kms-encryption-context) constructed above
```
seebees, line 823: I don't think you construct that above? But you have the branch key context, and you can use #encryption-context-from-authenticated-branch-key-context

framework/branch-key-store.md
```diff
@@ -870,36 +846,36 @@ A branch key record MUST include the following key-value pairs:
...
 - [Branch Key Id](./structures.md#branch-key-id) MUST be the `branch-key-id`
 - [Branch Key Version](./structures.md#branch-key-version)
   The version string MUST start with `branch:version:`.
   The remaining string encoded as UTF8 bytes MUST be the Branch Key version.
 - [Encryption Context](./structures.md#encryption-context-3) MUST be constructed by
-  [Custom Encryption Context From Authenticated Encryption Context](#custom-encryption-context-from-authenticated-encryption-context)
+  [Encryption Context From Authenticated Branch Key Context](#encryption-context-from-authenticated-branch-key-context)
 
-### Custom Encryption Context From Authenticated Encryption Context
+### Encryption Context From Authenticated Branch Key Context
```
seebees, line 871: I think that this is a really good section. I think you can use this everywhere.

seebees (commented):
Looking better. In create I would point to #encryption-context-from-authenticated-branch-key-context and use that to say "this encryption context". It can be an optimization that an implementation just uses the original input value.

framework/branch-key-store.md
rishav-karanjit, line 525:
```suggestion
    - Every key-value pair in the `encryption-context` and is associated with the branch key
```

framework/branch-key-store.md
rishav-karanjit, line 617:
```suggestion
    - Every key-value pair in the [Branch Key Context](./structures.md#branch-key-context) that is prepended with `aws-crypto-ec:`
```

framework/branch-key-store.md
rishav-karanjit, line 785: I like "sent by users" because for a first time reader its much easier to read.

rishav-karanjit: This is likely what you meant
```suggestion
If the `hierarchy-version` is v2, AWS KMS encryption context MUST be the [encryption context from authenticated branch key context](#encryption-context-from-authenticated-branch-key-context) without any transformation. This requirement applies regardless of the Branch Key Item's `type`.
```

framework/branch-key-store.md
rishav-karanjit, line 823:
```suggestion
- `EncryptionContext` MUST be the [encryption context from authenticated branch key context](#encryption-context-from-authenticated-branch-key-context) constructed above
```

framework/structures.md
rishav-karanjit, line 471:
```suggestion
The [encryption context](./structures.md#encryption-context) is a key-value pair provided by users and is associated with the beacon key.
```

framework/structures.md
rishav-karanjit: I need to remove this because technically it's not true. Will commit Tony's suggestion.

framework/branch-key-store.md
seebees: I don't think that a section above counts as "constructed above" :)
```suggestion
- `EncryptionContext` MUST be the [encryption context from authenticated branch key context](#encryption-context-from-authenticated-branch-key-context)
```

framework/branch-key-store.md
seebees: I think you want to reference the prefix. I would way something like all the additional pairs of the branch key context.

rishav-karanjit (commented):
I am not able to comment on code that does not have any diff but I am changing branch key context definition as below:
```
## Branch Key Context

Branch Key Context is a set of key value pairs (a Map) that contain contextual information about the Branch Key Item.

The Branch Key Context:

- MUST have a `branch-key-id` key who's value MUST not be an empty string
- MUST have a `type` key who's value MUST not be an empty string
- MUST have a `create-time` key with a value in ISO 8601 format in UTC
- MUST have a `tablename` key who's value is the logicalKeyStoreName
- MUST have a `kms-arn` key who's value is valid KMS ARN
- MUST have a `hierarchy-version` key who's value is either "1" or "2"
- MUST NOT have a `enc` key
- MAY have one or more keys prefixed with `aws-crypto-ec:` which is the encyption context send by the customer.
  The `aws-crypto-ec:` prefix is prepended by the library
- MUST NOT have any other keys apart from the ones mentioned above if `hierarchy-version` is "2"
```

framework/branch-key-store.md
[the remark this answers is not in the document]
seebees: There is some complexity here

framework/branch-key-store.md
```diff
@@ -601,8 +601,8 @@ List of TransactWriteItem:
     - “enc” (B): the wrapped DECRYPT_ONLY Branch Key `CiphertextBlob` from the KMS operation
     - “create-time” (S): `timestamp`
     - "kms-arn" (S): configured KMS Key
-    - “hierarchy-version” (N): 1
-    - Every key-value pair of the custom [encryption context](./structures.md#encryption-context-3) that is associated with the branch key
+    - “hierarchy-version” (N): either `1` or `2`, depending on the `hierarchy-version`
+    - Every key-value pair in the [encryption context](./structures.md#encryption-context) provided by users and is associated with the branch key
       MUST be added with an Attribute Name of `aws-crypto-ec:` + the Key and Attribute Value (S) of the value.
```
rishav-karanjit, line 606:
```suggestion
    - Every other key-value pair in the [branch key context](./structures.md#branch-key-context).
```

framework/branch-key-store.md
rishav-karanjit: We decided on a different suggestion.

framework/branch-key-store.md
```diff
@@ -817,13 +793,13 @@ this operation MUST call AWS DDB `GetItem`.
 Every attribute except for `enc` on the AWS DDB response item
 MUST be authenticated in the decryption of `enc`
 
-Every key in the constructed [encryption context](#encryption-context)
+Every key in the constructed [branch key context](#branch-key-context)
 except `tableName`
 MUST exist as a string attribute in the AWS DDB response item.
```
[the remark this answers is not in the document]
rishav-karanjit: Resolved here: [`3dadd20`](https://github.com/awslabs/aws-encryption-sdk-specification/pull/300/commits/3dadd20eecc2f42ea5ae31abde4382c95fdc244f#diff-c6735a6d874d8b9c52040541b81908877a6b9965b51d5b443c35af9895ec8676R797-R799)

PR #304, opened 2025-08-05
chore: add initial spec for S3 Encryption Client v3

kessplas:
_Issue #, if available:_

_Description of changes:_

See https://github.com/aws/amazon-s3-encryption-client-java/pull/481. This PR merges the spec added in that PR to the master branch of the spec repo. 

# Check any applicable:

### DIFF
--- a/s3-encryption/client.md
+++ b/s3-encryption/client.md
@@ -34,6 +34,7 @@ The S3EC provides client-side encryption for Amazon S3.
 
 The S3EC MUST adhere to the same interface for API operations as the conventional AWS SDK S3 client. 
 In other words, the SDK's conventional S3 client is able to be substituted for the S3EC. 
+The S3EC SHOULD support invoking operations unrelated to client-side encryption e.g. CopyObject as the conventional AWS SDK S3 client would.
 The S3EC MUST provide a different set of configuration options than the conventional S3 client.
 
 ## Configuration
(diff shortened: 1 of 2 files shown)


s3-encryption/client.md
```diff
@@ -0,0 +1,142 @@
...
+## Definitions
+
+### Conventions used in this document
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC2119](https://tools.ietf.org/html/rfc2119).
+
+### Key Material
+
+An AES or RSA key, or a valid AWS KMS key identifier.
```
josecorella, line 24: maybe a place for it here, but what format do we expect these in? just bytes? what format does the rsa public key need to be in? der/pem?

s3-encryption/client.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Client Initialization
+
+Client initialization refers to the process by which a new S3EC instance is created.
+Depending on the language, this could be a builder (using the Builder pattern) or a constructor.
```
josecorella, line 29: perhaps remove? afaik builder design pattern is primarily used for OOP - so maybe this is more of an implementation detail? 
```suggestion
```

s3-encryption/client.md
```diff
@@ -0,0 +1,142 @@
...
+When disabled, the S3EC MUST NOT decrypt objects encrypted using legacy content encryption algorithms; it MUST throw an exception when attempting to decrypt an object encrypted with a legacy content encryption algorithm.
+
+#### Enable Delayed Authentication
+
+The S3EC MUST support the option to enable or disable Delayed Authentication mode.
+Delayed Authentication mode MUST be set to false by default.
+When enabled, the S3EC MAY release plaintext from a stream which has not been authenticated.
+When disabled the S3EC MUST NOT release plaintext from a stream which has not been authenticated.
+
+#### Set Buffer Size
```
josecorella, line 80: what should happen if set buffer size is enabled but delayed auth mode is enabled? There is no behavior definition for this configuration? error? happy?

s3-encryption/client.md
```diff
@@ -0,0 +1,142 @@
...
+If the S3EC accepts any SDK client configuration options, then the S3EC should support all possible configuration options.
+
+### Other Dependencies
+
+#### Randomness
+
+The S3EC MAY accept a source of randomness during client initialization.
+The inclusion of a source of randomness is subject to language availability.
+
+## API Operations
```
josecorella, line 110: for readability i would separate this in two subheading sections
- MUST support:
  - [GetObject](link)
 - MAY support:
   - [CreateMultipartUpload](link)

s3-encryption/client.md
kessplas: good shout, i'm going to defer that to the specific keyrings.

s3-encryption/client.md
kessplas: sure, I tend to prefer being thorough and including conditional guidance, but you're right that this is an implementation detail. (Builder pattern may well go out of style eventually in OOP languages)

s3-encryption/client.md
kessplas: good shout, details added

josecorella (approved):

PR #321, opened 2026-07-09
feat(shim): Add generic shim and ESDK shim specifications

lucasmcdonald3:
## Summary

- **`shim.md`** — a generic specification for a shim library
- **`esdk-shim.md`** — the ESDK-specific shim specification


shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+- The shim MUST translate the target commitment policy to the core ESDK
+  commitment policy of the same meaning.
+- The shim MUST return an error when the target commitment policy is not a supported
+  value.
+
+### Encryption context
+
+Converting to the core ESDK:
+
+- The shim MUST translate the target's encryption context, given as a list of
```
kessplas, line 69: Do we want to get into encoding and such here?

lucasmcdonald3: No, the shim should translate from the target's type to the core library's type, and the core library does the encoding

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+An ESDK shim MUST conform to the generic [Shim Specification](./shim.md). This
+document defines only the ESDK-specific detail that specification defers to a
+core-library-specific spec: the concrete [type translations](#type-translation),
+[resources](#resources), and [operations](#operation-contracts) an ESDK shim
+exposes.
+
+A shim is not an ESDK implementation. It delegates the message format,
+cryptography, key derivation, commitment, and signing to the core ESDK.
+
+### Out of scope
```
josecorella, line 21: maybe stated below, but will test vectors run on the output produced by the shim? I assume so. Or are we expecting this layer of output to be no-op?

lucasmcdonald3: Test vectors will run on the "owned interface" which is the public interface, not on the "generated bindings" which are private

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+  to the core ESDK's "no limit" representation, and a positive value `n` to a
+  limit of `n`.
+
+### Frame length
+
+- The shim MUST translate the target frame length using the core ESDK's frame
+  length constructor, and MUST return an error when the core ESDK rejects the
+  value.
+
+### KMS configuration
```
josecorella, line 91: What about the KMS client? I don't know if you can translate one client from shim to core client, I assume the shim will expose a create_kms_client in the core client that allows for client config options to be set there as opposed to the shim layer?

josecorella: answered below i think

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+- The shim MUST translate a target algorithm suite identifier to the core ESDK
+  algorithm suite that has the same two-byte ID.
+- The shim MUST return an error when the target value is not one of the algorithm
+  suite IDs defined by the core ESDK.
+
+Converting from the core ESDK:
+
+- The shim MUST translate a core ESDK algorithm suite to the target algorithm
+  suite identifier that has the same two-byte ID.
+- The shim MUST return an error when the core ESDK produces an algorithm suite
```
ShubhamChaturvedi7, line 55: Isn't it contradictory in the sense that shim is missing one of the esdk types? Shouldn't the requirement be that shim MUST define all the values defined by the ESDK?

shim/esdk-shim.md
lucasmcdonald3: yup, asking the right questions, also see https://github.com/aws/aws-crypto-tools-rust/blob/aws-crypto-rust/aws-esdk-cpp-shim/esdk/shims/aws-esdk-cpp/src/lib.rs#L867-L878

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+per [Delegation](./shim.md#delegation).
+
+### Materials source
+
+- Each of `encrypt` and `decrypt` MUST be supplied with exactly one materials
+  source (see [Resources](#resources)).
+- The shim MUST return an error, and MUST NOT invoke the core ESDK, when no
+  materials source is supplied.
+
+### Encrypt inputs
```
josecorella, line 160: style nit;
since these are all musts, what is easier to duvet?
what we currently have or:
`encrypt`:
    -  MUST ...
    -  MUST ...
?

shim/esdk-shim.md
ShubhamChaturvedi7: I think what Kess meant is what if the target language uses utf8, but the core library expects utf16?

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Create hierarchical keyring
+
+- The shim MUST provide an operation that creates a hierarchical keyring backed
+  by the core ESDK.
+- Creating a hierarchical keyring MUST require a key store handle, checked as
+  defined in
+  [Shim Specification: Handles and lifetimes](./shim.md#handles-and-lifetimes),
+  and MUST provide it to the core ESDK.
+- Creating a hierarchical keyring MUST provide the target-supplied branch key id
```
josecorella, line 247: what about the branch key id supplier?

shim/esdk-shim.md
```diff
@@ -0,0 +1,256 @@
...
+
+The dependency edges are therefore: a key store depends on its service clients,
+and a materials source may depend on a key store.
+
+### Service client configuration
+
+The shim configures each AWS service client it creates from a client
+configuration supplied by the target. Any value the target supplies is passed
+through to the service client; where the target omits a value, the shim may
+apply a default.
```
ShubhamChaturvedi7: Again isn't this behavior goes against that shim may not interpret values?

shim/esdk-shim.md
lucasmcdonald3: Could drop `encrypt`, definitely easier to read

shim/esdk-shim.md
lucasmcdonald3: Let's talk about that it's tough, but we can add it later

shim/esdk-shim.md
lucasmcdonald3: The first requirement captures that, the second requirement just says "if the ESDK somehow spits out an unrecognized value, raise an error"

lucasmcdonald3: https://github.com/aws/aws-crypto-tools-rust/blob/aws-crypto-rust/aws-esdk-cpp-shim/esdk/shims/aws-esdk-cpp/src/lib.rs#L424-L449

shim/esdk-shim.md
lucasmcdonald3: Will write down that you need to write this in Rust and write down why etc.

shim/esdk-shim.md
lucasmcdonald3: I'll take a look at this but we should defer to core library

shim/esdk-shim.md
lucasmcdonald3: Yup, still just convert from target type to core type

josecorella (dismissed):

josecorella (approved):

shim/shim.md
```diff
@@ -146,35 +146,12 @@ The requirements below apply to every resource kind.
...
-- The shim MAY expose creation operations for only a subset of the
-  implementations the core library defines.
-- If the shim exposes a means of creating additional implementations, a handle
-  to such an implementation MUST be usable wherever that interface's handle
-  type is accepted.
+A core library can define interfaces that its consumer implements and that the
+core library invokes during an operation — for example, a custom keyring or a
+custom cryptographic materials manager.
+
+How a shim exposes such custom implementations is not yet specified; a future
```
josecorella, line 153: We may want to add that they are, "therefore unsupported" or do you think it is implied?

lucasmcdonald3: Eh I think it's implied here

